Issue #369, opened 2014-07-10
canRead calls are wasteful


benwerd:
They're repeated several times in the course of a page. I'd rather not do this, and only load or reload access groups when it's absolutely necessary.

PR #371, opened 2014-07-10
Caching interface (refs #369)

mapkyca:
This is the kind of thing I was thinking of Ben. 

There are two basic kinds of caches in this world, ones that are persistent outside of script execution and those that aren't (hence the two families). 

For now, databases should probably cache using the memory cache to do query caching. That said, since Idno uses PDO, it may be easier to cache at the object level. Again, use the memory cache for now, but since every object in the system has a unique ID it should be fairly straight forward to switch that over to a persistent cache of some sort.

Refs #369

### DIFF
--- a/Idno/Caching/Cache.php
+++ b/Idno/Caching/Cache.php
@@ -10,7 +10,7 @@ namespace Idno\Caching {
          * Store or replace a value in the cache.
          * 
          * @param $key string Identifier for this value
-         * @param $value string Value to store
+         * @param $value mixed Value to store
          * @return bool
          */
         abstract public function store($key, $value);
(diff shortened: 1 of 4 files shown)


benwerd:
Same question as the PR on the other pull request.

mapkyca:
I've not run the code for a while, since it seemed like you weren't interested and I needed to go on to build other stuff.

Mainly was for your reference...

mapkyca:
My suggestion, is have a look at this PR primarily - it just defines the interface and class hierachy: i.e. base cache class which is just a key store wrapper + two cache families defining a temporary cache and a persistent cache.

Temporary caches being ones that exist only for the lifetime of the script, and persistent ones that live on until they are invalidated (or expire).

This is a distinction I've used before and works quite well, since it lets you pick a suitable engine for e.g. query caches (which should be per script) and computed data or object caches which can persist for longer.

The code in #373 was intended as an illustration of usage (since this was easier done in code), but I didn't necessarily expect you to merge that one.

benwerd:
Rock on. Thank you.

PR #373, opened 2014-07-10
Caching getRecords

mapkyca:
All the patches in #371 plus an example usage caching getRecords() 

We may be able to do this cleaner by finding a better way to fingerprint PDO statements.

Adding this closes #371


benwerd:
I'm looking at this, trying to decide whether to finally bring it in. Have you been running this code without problems?

mapkyca:
Closing, as this was primarily an example

Issue #566, opened 2014-11-16
Following bookmarklet should scrape remote user photo


mapkyca:
##

mapkyca:
For clarification, following will show images when MF2 markup is found, however there is a privacy/leakage issue with including these verbatim. 

Same issue as https://www.marcus-povey.co.uk/2014/09/01/spying-on-a-site-using-webmention-and-mf2/, addressed in that instance by #431

PR #567, opened 2014-11-16
Fixing friends

mapkyca:
* Some cleanup of the add friend bookmark to handle missing photo (TODO: scrape remote photo #566)
* If no MF2 data is found, the user is given the opportunity to continue by filling in the details themselves.
* If a feed is detected, the user is fired down the subscribe workflow


mapkyca:
It'd be good to have a comment as to whether this is going to be merged - should be compatible, since it bounces to the subscription signup..

Issue #647, opened 2014-12-26
Invalid "Set site private" property


Phyks:
Hi,

On my Known instance, to make it public, I had to set "Set site private" to "Yes". I think the value of "Yes" and "No" in the template may be inverted with the db ones. My template is tabula rasa.

Thanks

mapkyca:
Confirmed, this is the wrong way round. Feel free to send a PR :)

Phyks:
@mapkyca done, thanks :)

PR #665, opened 2015-01-01
Fix issue #647

Phyks:
This is to fix the issue #647.

"Set site private" property was incorrect and should be "set site public". See the issue for full details.


benwerd:
Thank you!

PR #683, opened 2015-01-07
Load Google fonts from protocol-agnostic URL

### LINKED ISSUE #682: insecure content on /begin/register
When serving Known 0.6.5 over https, there is insecure content on the page /begin/register (coming from http://fonts.googleapis.com).

Grepping through the code, this seems to be coming from templates/default/shell/simple.tpl.php

michielbdejong:
I didn't test this, but I think something along these lines could fix #682


michielbdejong:
@Phyks I agree that would be even better

benwerd:
@Phyks Confirming that this is our plan. cc @erinjo 

In the meantime, I'm going to accept this PR. Thanks, @michielbdejong!

Phyks:
@benwerd nice !

PR #696, opened 2015-01-09
u-video u-audio for video audio posts respectively

tantek:
use u-video u-audio for video audio posts respectively per http://microformats.org/wiki/h-entry#Proposed_Additions. Fixes issue 694 and issue 695.


benwerd:
Oops. Thanks!

PR #703, opened 2015-01-10
Update requirements.rst

clintpatty:


benwerd:
This change makes it clear that you only need PDO if you're using MySQL as a back-end database. Thanks, @clintpatty!

PR #704, opened 2015-01-10
Update instructions.rst

clintpatty:


benwerd:
I'm going to merge this and then alter the description: in most cases, the installer _should_ create the .htaccess file for you.

clintpatty:
Hi Ben,

Is this worth noting, if this is the first you've heard of it? I don't want
to clutter it with what might have been specific to just my installation.

Clint

PR #706, opened 2015-01-10
name/caption of a photo post should be "p-name"

tantek:
name/caption of a photo post should be "p-name" not "p-photo". default.css changes should be made with this one.


tantek:
for issue 705

PR #707, opened 2015-01-10
remove p-photo rules, p-photo misuse replaced with p-name

tantek:
remove p-photo rules, p-photo misuse replaced with p-name, which already has rules styling it


tantek:
for issue 705

benwerd:
Rejecting this one, because p-name is too generic. I've added a new .photo-title class and changed the rules accordingly.

PR #709, opened 2015-01-10
remove p-photo rules, p-name rules handle it now

tantek:
remove p-photo rules, p-photo misuse replaced with p-name, which already has rules styling it


tantek:
for issue 705

PR #717, opened 2015-01-15
Known mentions

mapkyca:
Since the project no longer has the mentions library as a submodule, we can directly apply the patches submitted to the upstream project rather than wait.


benwerd:
Boom. You're a genius.

PR #720, opened 2015-01-18
Webmention receipt: accept more complex mf2 combinations

### LINKED ISSUE #701: Handle more complex mf2 combinations
@kylewm used this post:

https://kylewm.com/2015/01/we-are-admiring-your-podcast-in-indiechat-i-hope

To link to this post:

http://werd.io/2014/shoplifters-and-thieves-podcastsunday

But Known didn't handle it because the microformats boiled down to a p-in-reply-to with a bunch of objects inside it. It's worth going back and making sure we support the full set of expected indieweb variables.

kylewm:
Accept "u-in-reply-to h-cite" in addition to "u-in-reply-to" by checking 
each property value to see if it is a complex value, and extracting the
"url" property if it is.

@benwerd, this fixes #701 (tested on my partner's new Known site!).
Please don't be afraid to tell me if this is not how you would/want to do it.

### DIFF
--- a/Idno/Common/Entity.php
+++ b/Idno/Common/Entity.php
@@ -1384,6 +1384,28 @@
             }
 
             /**
+             * Many properties in mf2 can have either a simple string value or a complex
+             * object value, "u-in-reply-to h-cite" is a common example. This function
+             * takes a possibly mixed array, and returns an array of only strings.
+             *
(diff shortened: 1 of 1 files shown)


benwerd:
Rock *on*. Thank you!

PR #728, opened 2015-01-23
Add Uploads folder to gitignore

timmmmyboy:
Seams like something that should be in the default .gitignore


benwerd:
Yep. Thanks!

PR #762, opened 2015-02-15
Allow sitemaps.xml

mapkyca:


physcocode:
@mapkyca  I already tried this before , but this doesn't work . Google Webmaster tool is showing following error 

General HTTP error: HTTP 403 error (Forbidden)
HTTP Error: 403

So I ended up allowing all XML and causing a security threat !

maybe something is wrong in the 2.4 version of htaccess

mapkyca:
Have you tried applying the fix and then visiting the file directly e.g. https://your.site.com/sitemap.xml ?

Google likely doesn't check every visit, and also likely caches error responses.

PR #773, opened 2015-02-27
No password cache

mapkyca:
@benwerd This should fix it.


benwerd:
This was a great PR. Thank you!

PR #774, opened 2015-02-27
Diagnostics

mapkyca:
Introducing a diagnostics page stub.

It occurs to me that it might be handy for people on shared hosts, or anyone for that matter, to be able to get some data on the actual runtime config of their sites for diagnostic purposes.

Here's a stub, which collects some basic system info + has a hook for plugins to extend.

Later, it'd be cool to be able to execute unit tests from this page, but for now it just gives you a printout to refer to.


benwerd:
(Holding on this until the Bootstrap3 branch is merged)

mapkyca:
I updated this for BS3, plus made it button activated (like api key).

mapkyca:
Ok, diagnostics page much improved.

Now, you get a basic test on stuff warmup tests for (modules, upload directory), which is rendered in a friendly way to the user.

You can get some detailed diagnostics by clicking the button. 

Hopefully this will help folk who are having permission problems on upload directories etc after upgrade.

Issue #203, opened 2014-03-24
Harden Known against new attack landscape #discussion


mapkyca:
See #202... what attacks do we see in the post snowden landscape, what can we do to protect the citizen against them?

Starting off with a bit of a brainstorm, random thoughts in no particular order:
- Client -> server comms. Largely solved by TLS
- User private messages. Look at TLS for transport, give users public keys for private messages
- Protecting social graph. Encrypt friend list to the user(?). Database on encrypted drive. Recommendation to remove logs. Ability to route intersite comms over TLS and TOR (see #202)
- Cartel attack. Mitigated by distributed nature...? What else...?
- Updated 020150703 based on comment below: Work to remove unique identifiers from protocols (certainly non-secured protocols) 

More later, for now food, @benwerd when would be a good time to chat? my hands are tired from typing and it'd be good to kick this around for a bit :) 
##

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/idno/issues/203" class="u-in-reply-to"></a>Also, it seems, on airline wifi: https://nakedsecurity.sophos.com/2015/01/07/gogo-forges-youtube-ssl-certificate-to-throttle-high-bandwith-usage-on-flights/

mapkyca:
The IETF have an RFC (https://tools.ietf.org/html/rfc7258) which all internet protocols should factor. 

In a nutshell, things should be built with an understanding of selector based surveillance. Largely this means never sending unique identifiers in the clear.

In an ideal world I'd suggest not even using sessions on non TLS connections, but I'm aware that's not practical for the project, at least at this stage.

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/203" class="u-in-reply-to"></a>Bumping with the info that letsencrypt is now out. Is it possible to integrate with this somehow?

For example, could the installer detect whether it's installed and trigger a generation of cert?

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/203" class="u-in-reply-to"></a>Also noting that given that TLS is now essentially free, can we revisit the level of nudging and pestering we do?

Modern browsers now start giving in-your-face warnings when a password input field is delivered over a non-TLS connection, with the obvious intention of eventually blocking POSTing of those forms entirely.

Could we start warnings? Perhaps starting with admin logged in warnings on the diagnostics terminal and ramping up from there...?

[...]

PR #789, opened 2015-03-12
https in hub link

michielbdejong:
Haven't tested this, but would assume PuSH would work identically over https as over http, @julien51 can you confirm?


julien51:
Yes, of course!

benwerd:
Thank you!

mapkyca:
Nice spot. 

As a matter of policy I'm firmly of the opinion that everyone should default to https rather than http (ref #203 etc). At this point port 80 http can be considered deprecated, which is a view shared by the IETF.

PR #791, opened 2015-03-13
Removing duplicate headers

mapkyca:


benwerd:
Thank you as always :)

mapkyca:
Always happy to help :)

PR #794, opened 2015-03-16
Change to have pageTitle persist if entered

sgjudd:
The pageTitle was not being saved when entering or updating a Like. The code was checking if $title was empty, but not doing anything if it wasn't. I added an else statement to save it if it isn't empty.


benwerd:
Thank you!

PR #795, opened 2015-03-17
RIP Terry Pratchett

mapkyca:
As per https://np.reddit.com/r/discworld/comments/2yt9j6/gnu_terry_pratchett/

... and there's a Dogeon export template in core ffs.


benwerd:
Sold.

Issue #808, opened 2015-03-21
API Response is confusing


mapkyca:
Successful post returns the object, but also has some gumph like "action": "annotation/post"

e.g.

```
{
    "action": "annotation/post",
    "class": "stars",
    "confirm": false,
    "confirm-text": "Are you sure?",
    "data": {
        "object": "http://mapkycacom.icewing.dev.hub/2015/this-is-a-longer-test-some-other-stuff-which-should-3",
        "type": "like"
    },

...
```
##

mapkyca:
Additionally, GET on /status/edit for example produces a bunch of confusing data in fields, including a seemingly random "username" field, and a "service" field that seemingly is just the last value from the "services" array.

mapkyca:
Additionally "name": "marcus povey" on a GET to status/edit.

I can't quite establish where this is being generated, but I suspect its down to $vars being serialised by the json page shell. 

If this is true, then it's a bug in Bonita (benwerd/bonita#3), which causes variables to bleed across scope making issues like this common. I maintain that this is a recipe for disaster.

mapkyca:
Additionally, take end.tpl.php, which will output annotations etc.

In an API view, only the *last* set of annotations can ever be displayed.

Perhaps there needs to be a better way to generate non-html views other than rendering $vars....

mapkyca:
Referencing the annotations, I'm wondering if in this instance it's actually a spillover, perhaps the default object's json export should include annotations instead.

benwerd:
Is this still an issue do you think?

mapkyca:
Less so, but sometimes... I think theres a fundamental problem with variable scope in the bonita framework which causes lots of unintentional leakage

kylewm:
API responses also always have "messages", usually an empty list. this definitely confused me the first time I saw it.

mapkyca:
<a href="https://github.com/idno/Known/issues/808" class="u-in-reply-to"></a>I'd be tempted to unset the empty array, aye...

mapkyca:
Closing, as this is really a bonita issue.

PR #816, opened 2015-03-23
Tidy api result

### LINKED ISSUE #815: Utility API endpoints
Not sure how this'd fit in to the current page structure, but it'd be handy to have endpoints to query details of various things. 

For example, one I'd like right now is to be able to ask the site what syndication methods are available for a given content type.

Perhaps this could be accomplished via a GET to the appropriate xxxx/edit page? (Doing this currently produces something of what you'd want, it lists syndication methods but doesn't list corresponding accounts)

mapkyca:
Some work to prevent the worst cases of scope pollution in api views (#808) and closes #815 by returning more useful service account details in api result.

I think we may need to rethink how views are rendered, especially how data for the api is collected. Rendering $vars as json presents two problems 1) scope violation due to the long standing bug in bonita, 2) it's not possible to override views for an object because you can't just replace an object template and have it rendered.

### DIFF
--- a/templates/default/content/end.tpl.php
+++ b/templates/default/content/end.tpl.php
@@ -115,6 +115,8 @@
                             echo $this->__(array('annotations' => $mentions))->draw('entity/annotations/mentions');
                         }
 
+                        unset($this->vars['annotations']);
+                        unset($this->vars['annotation_permalink']);
                     ?>
 
                 </div>
--- a/templates/default/content/syndication.tpl.php
+++ b/templates/default/content/syndication.tpl.php
@@ -2,14 +2,23 @@
 
     $buttons = '';
     if (!empty($vars['services'])) {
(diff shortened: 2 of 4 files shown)


benwerd:
Why is annotations a leakage?

mapkyca:
Because it's unintended, and the results are not what you expect. Because variables aren't scoped, every variable of 'name' clobbers the value before it, meaning you'll only ever get the last annotation. Worse, since only some annotations have some fields you get values from one polluting another.

Better would be to have the object export annotations itself rather than have cruft collect in $vars, best would be to have variables explicitly passed to draw() as a parameter, or, if you're wedded to the passthru, have it push values onto a stack which is popped after a draw.

Either way, the value in $vars['annotation'] which is rendered in json is useless.

PR #820, opened 2015-03-25
Secure cookies

mapkyca:
Applying some of the recommendations from https://php.net/manual/en/session.security.php

Mainly, setting cookies_secure during secure sessions, but also, enabling strict mode sessions and switching the session ID algorithm from MD5 to sha256


mapkyca:
Two recommendations not addressed by this patch:

* Set HSTS on secure sessions
* Detect switch from HTTP to HTTPS and regenerate session ID when it happens

mapkyca:
One thing I notice is that you regenerate the session ID on every request. This isn't the end of the world (and does handle the HTTP->HTTPS switch), but may be overkill.

PR #826, opened 2015-03-28
Change empty object check

sgjudd:
entry/edit page title was not reflecting whether it is a new post or not. Changed the empty($object) check to check for empty id.


benwerd:
Thank you!

PR #828, opened 2015-03-28
Regenerate session ID on privilege change rather than every page load

mapkyca:
Seems to make sessions more stable.

I've tested with picture upload: before session was logged out after large picture upload, after patch session is preserved.

### DIFF
--- a/Idno/Core/Session.php
+++ b/Idno/Core/Session.php
@@ -32,7 +32,6 @@
                 session_name(site()->config->sessionname);
                 session_start();
                 session_cache_limiter('public');
-                session_regenerate_id(true);
 
                 // Session login / logout
                 site()->addPageHandler('/session/login', '\Idno\Pages\Session\Login', true);
@@ -373,6 +372,8 @@
             {
                 $return = $this->refreshSessionUser($user);
 
+                session_regenerate_id(true);
+                
                 return \Idno\Core\site()->triggerEvent('user/auth', array('user' => $user), $return);
             }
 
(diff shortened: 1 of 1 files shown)


benwerd:
Trying it, but may re-pull it if same issue persists.

mapkyca:
Fair enough.

My theory is that since session mode was being regenerated on every page load, the addition of strict mode meant that a client supplied (essentially client initiated) session was invalid and the session was destroyed. This is a *good thing*, since the client has no business supplying a session id in the header.

Basically, sessions relied on exact synchronisation, which meant it was very brittle if something, for example, took a long time, or hit a page via ajax, etc.

benwerd:
It's a good thing technically but a lousy thing in practice, because in the real world pages take an arbitrary amount of time to load. This explains why some people were being logged out on every photo upload.

This makes sense - the new fix works for me, and hopefully others can successfully test.

mapkyca:
It'd also break over dodgy connections, or if you click a link and stop the load or navigate away. Lots of reason this approach could fail :/

benwerd:
Seems to me that JavaScript callbacks would also break it? That makes it unacceptable for use in 99% of web apps.

mapkyca:
Depends what "it" is... strict mode is good, the client has no business setting a session ID. But, it exposes an issue in our code which is that session IDs were being regenerated every page load. 

The fact that sessions persisted before was basically because we were performing, in essence, a session fixation attack on ourselves O-o

benwerd:
Understood. Because we were _already_ regenerating IDs, the combination caused an issue. Agree about regenerating IDs less often.

kylewm:
This seems to fix the issue for me. tried uploading several large images and I'm no longer logged out.

PR #848, opened 2015-04-04
Hide API key on settings/tools when not viewed securely

mapkyca:


benwerd:
Warnings yes; disabling content no. Also this links to a URL that might well not exist and throw a browser error.

Filing under "not for now".

mapkyca:
Warning in this context wouldn't work, since the damage is already done! Maybe a js call to populate or something.

PR #865, opened 2015-04-24
Validate session

mapkyca:
Validates session against a privilege elevation attack on intercepted insecure cookies + acts as a stub for additional validation.

### DIFF
--- a/Idno/Core/Session.php
+++ b/Idno/Core/Session.php
@@ -33,18 +33,20 @@
                 session_start();
                 session_cache_limiter('public');
                 
+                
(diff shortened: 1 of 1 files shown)


benwerd:
Ooh. We should have talked about this in our call. Can you talk me through what this does and why?

mapkyca:
So, general best practice is to destroy the session should anything go screwy (hence validate to be extended). A cookie session used in an insecure session can not be trusted, especially persistent, so recommended practice is to destroy sessions when http sessions are used for https to preserve the integrity of a tls session.

Issue #858, opened 2015-04-21
404 on /begin with defalt Apache config AllowOverride off


kylewm:
Ubuntu configuration has "AllowOverride off" for /var/www which disallows .htaccess from rewriting urls, which means that lots of users hit a 404 immediately after installation succeeds.

It's an [easy enough fix](http://indiewebcamp.com/Known#Fix_404_Not_Found_on_.2Fbegin) but it seems to bite lots of people (informal observation based on questions in #knownchat).

If it's possible to detect this error during installation and give a helpful warning, I think that would be ideal.

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/858" class="u-in-reply-to"></a>There's no good way to detect for this *specifically*, but perhaps a test to see if /begin/register or similar returns a 404 on warmup

benwerd:
An obvious solution would be to try and load a piece of JavaScript that
lives at a rewritten URL as part of the /warmup process. Or at least,
that's what I've been thinking.

On Thu, Apr 23, 2015 at 6:07 AM, Marcus Povey <notifications@github.com>
wrote:

> <https://github.com/idno/Known/issues/858>There's no good way to detect
> for this *specifically*, but perhaps a test to see if /begin/register or
> similar returns a 404 on warmup
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/idno/Known/issues/858#issuecomment-95578046>.
>



-- 
Ben Werdmuller
<http://goog_1933028737>
benwerd.com | werd.io

+1 (312) 488-9373

kylewm:
You're thinking a JS file (or some other static resource) instead of /being/register to avoid false positives from other php configuration errors?

benwerd:
Yeah, because then it could actually reliably set the prerequisite page -
the installer could fail to continue if you don't have it set up.

On Thursday, April 23, 2015, Kyle Mahan <notifications@github.com> wrote:

> You're thinking a JS file (or some other static resource) instead of
> /being/register to avoid false positives from other php configuration
> errors?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/idno/Known/issues/858#issuecomment-95644536>.
>


-- 
Ben Werdmuller
<http://goog_1933028737>
benwerd.com | werd.io

+1 (312) 488-9373

kylewm:
[...]

PR #867, opened 2015-04-25
Detect "AllowOverride None" during setup

kylewm:
Use a silly redirect in .htaccess (/js/canary.js -> /js/default.js),
and use it in warmup/settings.php to detect whether .htaccess is
allowed to rewrite URLs. Hoping this helps fewer self-hosters
get hung up on this common configuration glitch!

This is real dog-sciencey as I fumble through learning php;
please feel free to reject or ask for changes/improvements.

Addresses #858


benwerd:
Yay, thank you!

PR #868, opened 2015-04-25
Warmup ssl

mapkyca:
Place a message during warmup/requirements.php to alert the administrator when they're installing Known on a non-secure site, nudging them to consider installing a certificate.

This is only a warning at the moment, but there's a switch for making it a requirement at some point in the future when port 80 is more widely deprecated.

### DIFF
--- a/css/simple.css
+++ b/css/simple.css
@@ -907,6 +907,10 @@ a.btn:hover {
     background-color: #FAC9B9;
 }
 
+.warning {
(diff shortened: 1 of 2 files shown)


physcocode:
Showing an error is a good idea  , but forcing them to SSL is not a great idea cause many people are intrested in just using CMS and not seeting up big user base or a secure server . forcing SSL may cause known to loose market !

PR #871, opened 2015-04-26
Make curl settings more configurable

mapkyca:
Among other things this makes known now fully support tor hidden services


physcocode:
A double +1 for this *_*

Issue #843, opened 2015-04-02
Bridgy connect UX


beck24:
Just connected Bridgy to both twitter and facebook.  When you return to the Bridgy page after authorizing the app there's no message or anything indicating it was successful and the page looks the same with call-to-action buttons for activating the services you've just activated.

I checked on Twitter and FB and the Bridgy app is authorized.  It would be better if it would tell you that it's already connected.

![screen shot 2015-04-02 at 10 52 36 am](https://cloud.githubusercontent.com/assets/738363/6973659/134bc3aa-d941-11e4-8f4d-5ca75cc512d3.png)

erinjo:
Completely agree. We've talked a lot about this. There actually isn't anything passed back to us from Bridgy, so we don't have a way to tell if people are authenticated or not. In a perfect world, we would also know if they were no longer authenticated with Bridgy, since they could also remove their connection from the Bridgy website. We've been discussing possible hacks and ways that we could capture some of this information, with the short term goal of making it feel similar to authenticating with other social networks and the long term goal of just building it in completely so that the user never has to authenticate through Bridgy.

beck24:
Fair enough, that does certainly make it difficult.  I'll just leave this ticket here as a reminder then in case someone comes up with a good solution.

andtheWings:
+1 Yeah, I found this pretty confusing although I understand the limitations you're facing.

kylewm:
What is needed on the Bridgy side to support this? More information when Bridgy redirects back to Known, or a query endpoint that you can use to check the status of an account every time they open this page?

benwerd:
A query endpoint would probably be the least brittle. But it also depends
on what's easy from your side?

On Tue, Apr 28, 2015 at 10:50 AM, Kyle Mahan <notifications@github.com>
wrote:

> What is needed on the Bridgy side to support this? More information when
> Bridgy redirects back to Known, or a query endpoint that you can use to
> check the status of an account every time they open this page?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/idno/Known/issues/843#issuecomment-97151871>.
>



-- 
Ben Werdmuller
<http://goog_1933028737>
benwerd.com | werd.io

+1 (312) 488-9373

kylewm:
I think both would be pretty easy, I'll work on an endpoint and let you know!

benwerd:
Awesome! Let me know if I can help in any way.

[...]

PR #883, opened 2015-05-07
store Bridgy connected/disconnected status

kylewm:
- better visual feedback when connected
- allow users to disconnect via the same interface

this addresses issue #843 

still TODO:

- Refetch and parse the Bridgy user page to update connected status
  in the future.


benwerd:
Amazing!! Thank you thank you thank you.

PR #896, opened 2015-05-14
re-check bridgy connection status

kylewm:
- First call checks whether account enabledness has changed
- Re-render the account page and have JS swap out the relevant section

This is the last thing I was planning to do for #843

Of possible minor interest, I wanted to re-render the Account page as part of the Check handler (so that there would only need to be one request). So I would update $user, and then draw the Account page, but it never seemed to pick up the changes to user... guessing it has something to do with the way the session is persisted. Rendering in a second JS request proved much more reliable.


benwerd:
Awesome!

PR #911, opened 2015-05-28
Add optional field to posse links to store bare item ID

misuba:
This enhancement will open up the possibility of doing more in plugins with major social-network APIs.

### DIFF
--- a/Idno/Common/Entity.php
+++ b/Idno/Common/Entity.php
@@ -978,14 +978,14 @@
              * @param $url
              * @return bool
              */
-            function setPosseLink($service, $url, $identifier = '')
+            function setPosseLink($service, $url, $identifier = '', $item_id = '')
             {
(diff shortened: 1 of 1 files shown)


kylewm:
Just out of curiosity, what do you want to do with the bare ID that cannot be done with the syndication URL?

misuba:
Make some other calls to the Twitter API from the Twitter plugin (or another plugin, I'm waffling on that). It would be nice not to have to fish IDs out of URLs, especially if the URL format ever changes.

mapkyca:
There should be nothing preventing a plugin from listening events and adding this info of course.

misuba:
...besides the total lack of documentation on events. :-)

I see the existing Twitter plugin calling the posse methods; that's all I'm going by.

benwerd:
Agree on:

* documentation (we really need a concerted effort to improve this)
* using IDs over URIs

PR #916, opened 2015-05-31
Fixing bookmarks

mapkyca:
Bad code no longer breaks page, but we still need to work out why 

http://nymag.com/scienceofus/2015/05/some-people-get-sad-in-the-summer.html 

Breaks title extraction regex in callback

### DIFF
--- a/IdnoPlugins/Like/templates/default/entity/Like.tpl.php
+++ b/IdnoPlugins/Like/templates/default/entity/Like.tpl.php
@@ -13,7 +13,7 @@
 
 ?>
 <div class="">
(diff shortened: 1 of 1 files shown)


benwerd:
That was fast! I was just in the process of writing the exact same thing.

mapkyca:
Happy to help! I think the problem was first caused by a bug in the title extract regex.

benwerd:
It's an xpath query, isn't it? *goes to check code*

mapkyca:
There's a callback page, and it looks like the regex is not finding the title close tag and so is going off the reservation.

mapkyca:
Got it... you need to invert the greedyness - basically, the sample page had multiple <title></title> tags in the body (naughty!), so broke the title extraction regex.

PR #926, opened 2015-06-08
use "u-uid u-url" to find wm permalink

kylewm:
If it's specified "u-uid u-url" gives the post permalink,
as opposed to "u-url" by itself which can include things
like shortlinks.

Known currently concatenates multiple u-urls together
which btw, is probably a bad idea in general
bc you should always be getting full absolute URLs as
the value of u- properties. (so concatenating them
would always give an incorrect URL)


kylewm:
leo

benwerd:
:ok_hand:

PR #936, opened 2015-06-16
Xcache pages

mapkyca:
* Adds XCache key cache support (with stub for other mechanisms) - useful alternative to database storage for cached data
* Cache 304 data on object and file views, so database is no longer required for 304 test


mapkyca:
You may just want to merge the xcache portion, it's up to you. Var cache is handy to store the result of calculated values of expensive calculations, that said there aren't that many of those just yet ;)

PR #953, opened 2015-06-30
Modify soundcloud embed

mapkyca:
Merges code from my soundcloud plugin which loads the embed asynchronously via javascript oembed call rather than blocking with file_get_contents().


danito:
As we're all using Known, @benwerd should post a status on the withknown stream and we reply (webmention) from our own Known with the link to our plugins... 
;-)

mapkyca:
That could work. I also started a stub for a github project plugin (https://github.com/mapkyca/KnownGithubProjects) a while back. The idea is that lets you link a github entry to a github project and have it automatically post updates... might be of use adapting in this context, but I've not really had time to do much dev on it.

PR #955, opened 2015-07-02
Removing errant html file

mapkyca:


benwerd:
Whoops! Thanks for the catch.

Issue #978, opened 2015-07-22
Mongo and Mysql data sources have inconsistent behaviour


mapkyca:
getAnyRecord return Array in mongo, completed PHP object of correct class in MySQL

benwerd:
This is odd. By my reading, MySQL getRecord should also return an array - it's just a wrapped call to PDO, with associative array specified. Could the issue have been something else?

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/978#issuecomment-125421355" class="u-in-reply-to"></a>Check out the dataconcierge test class for an example of what I mean (had to put in an exception to handle it).

Also, for Mongo, getID returns an object which must be explicitly cast into a string to get a string value.

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/978" class="u-in-reply-to"></a>Looking again I think I was wrong about getRecord, but getAnyRecord is definitely inconsistent:

On Mysql (and sqlite which reuses a lot of code), getAnyRecord passes the pdo row through rowToEntity. Mongo returns findOne, which returns an array.

kylewm:
There are two issues:

kylewm:
(Thank goodness for unit tests)

Started trying to fix this and discovered DataConcierge::getRecordByUUID, getRecords, getAnyRecord return exactly the array that was saved in saveRecord. MySQL::getRecord gets the columns of the DB instead. The array that was saved in saveArray is encoded as json in the 'contents' field:

```
array (size=7)
  'uuid' => string 'http://known.dev/profile/kylewm' (length=31)
  '_id' => string 'fad2bc72f24adfd7c8ade05cbe2d681c' (length=32)
  'owner' => string '' (length=0)
  'entity_subtype' => string 'Idno\Entities\User' (length=18)
  'created' => string '2016-01-29 18:27:46' (length=19)
  'contents' => string '{"access":"PUBLIC","owner":"","email":"kyle@kylewm.com",... (length=821)
  'search' => string 'kyle kylewm  ' (length=13)
```

I'd like to change these methods in MySQL (and friends) to return the same array DataConcierge returns. The cool thing there is I think I could get rid of their implementations of rowToEntity and just use the one in DataConcierge.

mapkyca:
[...]

PR #980, opened 2015-07-22
Adds SQLite3 support

mapkyca:
This patch does a bunch (below) but primarily it introduces support for SQLite file based sql storage.

I did this because SQL/Mongo has quite a lot of overhead in terms of both complexity and resources compared to straight files, and often seemed overkill for folk running Known on low powered devices (embedded / single use sites / simple hosting etc).

It provides a nice fallback, a useful tool for quickly building test builds, as well as reduces some of the admin overhead for folk.

I've got plans for this for some IoT type stuff, and was meaning to do it for a while, but it was prompted by some guy on IRC who's SD card bought the farm, and I figured a SQLite db might be more appropriate.

Anyway, this patch does:

* Introduces SQLite dataconcierge 
* Adds SQLite to travis builds
* Introduces a dataconciege test to build which tests mysql, sqlite and mongo builds
* Fixes a few bugs exposed by these unit tests
* Exposes some inconsistencies between mongo and mysql (#978)

Anyway, it seems to be fairly stable, but should probably be considered "experimental" until the path gets trodden a little more.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,9 @@ env:
   - DB=mongo
   - DB=sqlite3
 
-#branches:
-#  only:
-#    - master
+branches:
+  only:
+    - master
 
 before_script:
   # Update and install base system
@@ -50,8 +50,6 @@ after_failure:
   - sudo cat $TRAVIS_BUILD_DIR/apache-error.log
 
 notifications:
-   email:
-     - marcus@dushka.co.uk
-#  email:
-#    - hello@withknown.com
-#  irc: "irc.freenode.net#knownchat"
+  email:
+    - hello@withknown.com
+  irc: "irc.freenode.net#knownchat"
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,7 @@ services:
 
 after_failure:
   - sudo cat $TRAVIS_BUILD_DIR/apache-error.log
+  - mysql -utravis -e "use known_unittest; select * from entities where match (search) against ('test sear');"
 
 notifications:
    email:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ php:
 env:
   - DB=mysql
   - DB=mongo
+  - DB=sqlite3
 
 branches:
   only:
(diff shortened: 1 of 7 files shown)


benwerd:
Waiting for Travis and then I'll merge this in. Your commit messages are ... interesting ;)

kylewm:
oh hell yeah! this will also be useful for people who want to run on Sandstorm.io (since everything is containerized, they recommend SQLite if possible)

mapkyca:
Frustration: got bit on the bum by mysql's weird FTS restraints

paulproteus:
My main question here is -- what happens if two simultaneous HTTP requests to Known want to writes? Will these be properly sequenced? Will one of them fail with Error 500?

If things will be honky-dory, then awesome. But I just wanted to make sure to check what the answer to this is.

mapkyca:
SQLite uses a file on disk, so it should be (almost) no different than accessing any other static file (for reads). Concurrent writes you'll start to lower performance since you enter the standard lock & queue IO gubbins... how bad that is depends on your file system.

paulproteus:
Right, I guess I just want to make sure someone has tested that this is
configured in a way that doesn't cause interleaved database transactions,
and doesn't cause error 500.

mapkyca:
It's implemented using PHP's PDO interface, and sqlite purports to be ACID, so I'm going to kick that one down the road. ;)

Phyks:
Do you plan on adding an easy way to migrate from a MySQL / MongoDB database to an SQLite one for old users?

benwerd:
There's going to be better site to site migration, and I suppose you could
use that for this.

On Sunday, July 26, 2015, Lucas Verney <notifications@github.com> wrote:

> Do you plan on adding an easy way to migrate from a MySQL / MongoDB
> database to an SQLite one for old users?
>
> —
> <https://github.com/idno/Known/pull/980#issuecomment-125033518>.
>


-- 
Ben Werdmuller
<http://goog_1933028737>
benwerd.com | werd.io

+1 (312) 488-9373

mapkyca:
Also, SQLite is an "as well as" rather than "instead of" wrt MySQL/mongo... Hopefully useful, but not a replacement :)

Phyks:
Ok :) I was trying to get rid of MySQL on my server as very few apps need it :)

kylewm:
@mapkyca could you elaborate on why it wouldn't be useful as a replacement?

mapkyca:
It certainly would be a useful replacement for some people, I'm just saying that currently mongo is the most mature and MySQL is the current "blessed" engine, so both of those are going to be much more well trodden paths, and I don't thing @benwerd has plans on changing the defaults any time soon :)

FWIW: Sqlite is the engine I'm using in my dev environment, as it lets me swap configs and scenarios around very quickly.

PR #983, opened 2015-07-27
Fixed font-awesome class for github icon

jonaswisplinghoff:
Github icon needs class `fa` instead of `a` to be a font awesome icon. Was just a typo I guess.


benwerd:
Thank you!

PR #990, opened 2015-07-29
move Core/Time to Common/Time

kylewm:
Discrepancy in that Core/Time.php says it's in Common/Time.

Figured I'd send a PR to save you half a minute in case Common/Time.php is the right one.


benwerd:
Indeed! It actually should be Core/Time, but thanks so much for the patch! I've committed a fix.

PR #991, opened 2015-07-29
Postgres support

### LINKED ISSUE #76: Postgres support
It seems that Idno is designed to use different data stores using this class: https://github.com/idno/idno/blob/master/Idno/Core/DataConcierge.php, is that correct? If so, it should be possible to swap that out with a wrapper that stores things in Postgres. Do you see any issues with that? I would be more comfortable running this under Postgres or MySQL to avoid introducing a new piece of technology on my stack.

(http://aaron.pk/r4Rq3)

mapkyca:
Closes #76 

This is very much experimental, because Postgres was giving me a headache, but I'm like a dog with a bone with this sort of thing ;)

Things not working quite yet:

* Full text search
* A better way to do on delete insertions (not supported by postgres)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_script:
   - sudo sed -e "s?%TRAVIS_BUILD_DIR%?$(pwd)?g" --in-place /etc/apache2/sites-available/default
   # Configure Known
   - if [[ "$DB" == "pgsql" ]]; then echo "PROVISIONING WITH PostgreSQL"; fi
-  - if [[ "$DB" == "mysql" ]]; then sudo cp -f Tests/build/config_pgsql.ini $TRAVIS_BUILD_DIR/config.ini; fi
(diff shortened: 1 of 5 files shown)


aaronpk:
What do you mean "on delete insertions"? Postgres supports triggers which let you do some pretty fancy stuff.

mapkyca:
Probably, but I didn't really have time to figure out the various UPSERT syntaxes ;) Someone who's a bit more of a postgres wizz will be able to make this much better I'm sure! But hopefully this is a useful starting point ...

kylewm:
ah, I see MySQL's `insert ... on duplicate key update ...` is not supported by Postgres. It looks like version 9.5 will introduce `insert ... on conflict (_id) do update ...` but it'll be quite a while until that's in most server distros.

it looks like the recommended way to make atomic upserts in older versions is with PLSQL and an exception handler, yuck. http://www.postgresql.org/docs/current/static/plpgsql-control-structures.html#PLPGSQL-UPSERT-EXAMPLE

aaronpk:
Yeah or just do it in code to avoid the problem.

kylewm:
Yeah, that is what @mapkyca did

PR #993, opened 2015-07-31
Postgres equivalent of MySQL "on duplicate update"

kylewm:
The Postgres equiv of MySQL's on duplicate update is kind of a weird
hack, but it seems to work.

Also made the hostname optional so you can connect to postgres
over a unix domain socket instead of a port.


mapkyca:
Nicely done!

benwerd:
Awesome!

PR #1005, opened 2015-08-07
Introducing the concept of logging contexts.

mapkyca:
Purpose: to allow certain submodules to set a context (or multiple contexts), so as to better see where log messages come from.


benwerd:
I like.

PR #1007, opened 2015-08-11
Limit some potential exfiltration exploits

mapkyca:


mapkyca:
Strongly recommend benwerd/bonita#11 as well...

PR #1010, opened 2015-08-11
Closes #1008: Validate form tokens on write actions, as per function …

### LINKED ISSUE #1008: POST fields should validate form actions
Forms have XSS tokens on the form, but these don't appear to be being checked when that form is submitted, correct me if I'm missing something...

mapkyca:
…specification in comments


mapkyca:
Ignore me, they were being validated, but the problem was a separate issue.

PR #1011, opened 2015-08-12
Updating Bonita to latest

### LINKED ISSUE #1009: Ability to bypass form validation
* Use validateToken() in your form with either () or ('/foo/bar/') in the validate parameter
* If Token is supplied BUT WRONG, form fails to validate (correct)
* If Token is MISSING, form will validate as correct (ERROR)

XSS is therefore possible since all you have to do is not send those values, or strip them from any form being sent.

mapkyca:
Closes #1009


benwerd:
Thank you sir

mapkyca:
Here to help :)

PR #1015, opened 2015-08-16
Bridgy interactions bugfixes

kylewm:
- don't require users to put their Known url in their Twitter/Facebook profile.
  Send the user url along with the authorization request (I thought we had
  already done this??)
- fix typo on delete Facebook form action -- should be delete/start instead of
  facebook/start

I don't think twitter.tpl.php and facebook.tpl.php are used anywhere?


benwerd:
Oh, that's weird. I thought the URL was being sent too. Thanks for writing this.

benwerd:
I've just deployed the fix to withknown.com, too.

Issue #989, opened 2015-07-28
Reorder pages and categories in StaticPages [$50]


benwerd:
You should be able to re-order pages and categories using drag and drop from the StaticPages admin screen. This should then affect the display of links to these pages in the navigation bar.

cleverdevil:
+1 for this feature! Was just asking about it in IRC today.

benwerd:
Fixed by @enebe-nb! Thank you for your contribution!

markwaters:
How do I reorder static pages ?
I cannot see any controls apart from 'Make Homepage'

mapkyca:
<a href="https://github.com/idno/Known/issues/989" class="u-in-reply-to"></a>You can drag and drop your pages to reorder them on the page management page

markwaters:
Ahhaaa!
Got it!
Thanks @mapkyca , @benwerd and @enebe-nb

PR #1016, opened 2015-08-19
Add Reordering for StaticPages plugin

enebe-nb:
Add reordering using drag and drop as asked in #989 .
Uses https://github.com/voidberg/html5sortable as helper library.
Changes in categories are instantly reflected on pages table, but the changes on header bar are only updated on next navigation. (Just to simplify the code and don't touch on themes)
A new "priority" parameter was created on StaticPage entity, it is first set on first reordering.
Reorder on pages only work on same Category. (still have to edit the page to change its category).

I didn't found a way to include the JS library only for one page using the Page Class, so i included it on template.


enebe-nb:
Oh, tests i didn't saw them. Soon i'll send them.

mapkyca:
There aren't any at the moment, but there's support for them. If you want to add plugin specific tests (feel free to do so), you should create them in a directory called "Tests" inside your plugin, and extend the KnownTestCase class inside your WhateverTest class.

enebe-nb:
If i add tests for this plugin then i have to add for the whole plugin, also the reordering was implemented as ajax and that makes a little more hard to do automated tests, for example, using Selenium Framework.
I made the manual tests, but make an automated will take more time then the pull request already took. Please, see case you like this code, and what it needs more.

benwerd:
Merged. Thank you!

Issue #1018, opened 2015-08-20
Problem with pagination on mongo


mapkyca:
I mentioned this in an IRC chat, but if you've got a large number of entries in the database the current data functions will fail (even with pagination limits) because of the need for chronological sorting.

Not sure if this is a data concierge problem or a mongo problem, or both.

mapkyca:
Pretty sure this is down to the skip/sort/limits on the collection. Basically mongo will retrieve and sort the *entire* db before it applies limit and skip.

PR #1019, opened 2015-08-20
Export debug

mapkyca:
* Addresses problems exporting MongoDB ids
* Handles large exports better (however, there is still a problem with #1018)

### DIFF
--- a/Idno/Core/Migration.php
+++ b/Idno/Core/Migration.php
@@ -135,6 +135,7 @@
                         }
                     }
                     
+                    $results = null;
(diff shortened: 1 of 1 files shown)


misuba:
Do you still need to have outgoing mail configured to get an export?

mapkyca:
Wouldn't have thought so, but you obviously won't get notifications if not.

misuba:
I guess it depends what one means by "get an export." I did find it in /tmp. Now to get it to be useful...

PR #1038, opened 2015-09-01
We should return denied content, rather than setting 403 and proceeding anyway

mapkyca:


mapkyca:
Bumping, as I'll be doing a bunch of api work soon

PR #1077, opened 2015-10-06
Add ability to specify custom database handlers, outside of the exist…

mapkyca:
…ing Idno\Data\* tree.

### DIFF
--- a/Idno/Core/Idno.php
+++ b/Idno/Core/Idno.php
@@ -65,6 +65,10 @@
                             $db       = "Idno\\Data\\{$this->config->database}";
                             $this->db = new $db();
                         }
+                        if (empty($this->db) && class_exists("{$this->config->database}")) {
+                            $db       = "{$this->config->database}";
+                            $this->db = new $db();
+                        }
                         if (empty($this->db)) {
                             $this->db = new DataConcierge();
                         }


physcocode:
But why do we need to break conventions and declare the database class anywhere else ?

mapkyca:
Because you might want to have a plugin provide a data store that is not part of Known core.

physcocode:
That sounds intresting !

kylewm:
On IRC we ran into a problem with this -- @misuba had the database set to "MongoDB", the class `Idno\Data\MongoDB` doesn't exist, so it tries to instantiate the class new MongoDB() itself, which is not what we want!

Not sure the best way to fix, perhaps check both cases in the cases?

```php
                     case 'mongodb':
                     case 'MongoDB':
                         $this->db = new DataConcierge();
                         break;
```

mapkyca:
Not related to this patch, but worth maybe a strtolower on the switch clause...

kylewm:
@mapkyca  it's related to this patch because the value "MongoDB" is being caught by the new conditional:
```
 if (empty($this->db) && class_exists("{$this->config->database}")) {
```
and causing problems.

mapkyca:
Ahh, I see. Sorry, I misunderstood.

If we strtolower the switch this should stop the specific issue, but I think that maybe we should be testing instanceof after we create the new class and ditch it if it's not a subclass of the DataConcierge.

kylewm:
Does PHP have the ability to check that without instantiating the class? (In Java it'd be `DataConcierge.class.isAssignableFrom(...)`). In this particular case MongoDB didn't have a no-arg constructor, so just calling the constructor was failing

mapkyca:
Hmm... no, I'm not entirely sure it can, and instantiating a random class could be potentially problematic (although if you're exposed enough for this to occur, then you have bigger problems). Perhaps it'd be enough to check that the namespace contains either Idno or IdnoPlugins...? The goal of the patch was to allow plugins to provide their own DataConcierges without needing them be accepted into core.

benwerd:
PHP actually can, using is_subclass_of, which can take an object _or_ a class name.

mapkyca:
Shiny.

Issue #753, opened 2015-02-09
twitter avatar issues


srcr:
In my Know installation at home I have an issue with twitter avatar pictures.
when I download a picture from my site it will show me the following content:

<html><body>You are being <a 
href="https://pbs.twimg.com/profile_images/562501973593714688/5nxxcm2n.png">redirected</a>.</body></html>

Looks like there is an issue with the download process.

benwerd:
Uh oh. Is your site public? Can you point to a page where this is happening? (Are you using brid.gy to get the avatars?)

davidmead:
I'm seeing the same thing as @srcr.

Looking at the comments left on http://blog.davidjohnmead.com/2015/this-is-why-i-love-indieweb-starring-microformats#comments you can see the gravatars from peoples sites are coming through, but not the ones from Twitter.

![screen shot 2015-02-10 at 8 15 52 am](https://cloud.githubusercontent.com/assets/1266930/6127816/4c4c49d4-b0fd-11e4-868c-2ffe4530acda.png)

When right-clicking on Aaron's broken image I get this URL `http://blog.davidjohnmead.com/file/eb3363dd5d68c3848a72bdc84a97cdc4 ` , which entered into the browser address bar downloads a file. Launching that file gives me a redirect link.
 
![screen shot 2015-02-10 at 8 20 25 am](https://cloud.githubusercontent.com/assets/1266930/6127857/bc9b0a86-b0fd-11e4-9b8f-f03eb4052aa3.png)

Clicking on that, gives me this link `https://pbs.twimg.com/profile_images/3657148842/934fb225b84b8fd3effe5ab95bb18005.jpeg` which is Aaron's photo.

Not sure what's going on.

srcr:
@benwerd Indeed the problem that @davidmead describes is the same as mine, only var better described +1 My sites is in my profile, but In the meantime I downloaded the profile pictures manually so the issues is currently not visible on my site.

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/753" class="u-in-reply-to"></a>Looks very much like twitter has changed the URL of the image CDN, but not changed the url extracted by the MF2 parser. Alternatively, it's a bridgy thing, and it's not correctly getting the _true_ url from whatever shim it uses to turn twitter permalinks into MF2 marked up stuff.

In any case, we're attempting to pull the icon and hitting a redirect.

[...]

PR #1096, opened 2015-11-08
Follow redirections on file_get_contents()

paulcmal:
Quickfix for #753.

Adds a block to Webservice.php:file_get_contents() to check for redirections. Stops after 3 redirections to prevent redirect loops.

### DIFF
--- a/Idno/Core/Webservice.php
+++ b/Idno/Core/Webservice.php
@@ -283,6 +283,19 @@
             {
                 $result = self::get($url);
 
+                // Checking for redirects (HTTP codes 301 and 302)
+                $redirect_count = 0;
+				while (($result['response'] == 302) || ($result['response'] == 301)) {
+				    $redirect_count += 1;
+				    if ($redirect_count >= 3) {
+				        // We have followed 3 redirections already…
+				        // This may be a redirect loop so we'd better drop it already.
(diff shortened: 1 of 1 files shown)


benwerd:
Thank you!

paulcmal:
You're welcome!

> Stops after 3 redirections to prevent redirect loops.

As a matter of fact, it stops after *2* redirections. I have no idea if that's something we want to change, leave for the user to change from the admin interface, or just leave like this.

paulcmal:
Just so that we don't lose track.

I made two workarounds, 7eea1d5265782502968e2ef179b45d0cc852861c working with cURL params, ad0ab518eeb3d0e58c9cf958e82921783456c299 adding checks to file_get_contents().

Both work on my Known despite having an open_basedir. I found very little information on the inner workings of interactions between cURL and and open_basedir, but [this post on SO](https://stackoverflow.com/a/19675459) seems to suggest that open_basedir restrictions would affect incoming requests only.

I am running nginx 1.8.0, PHP 5.6.15 (with php-fpm) and curl 7.45.0.

Could somebody else on a different setup try 7eea1d5265782502968e2ef179b45d0cc852861c and see if it works despite open_basedir?

PR #1107, opened 2015-12-03
Cleaner status imports from WP

obra:
When a post coming from wordpress is less than 500 characters and has no title, treat it as a status update, rather than a full-on post.


benwerd:
Yay, thank you!

PR #1112, opened 2015-12-16
Minor point, but since this is a single use function that won't be ca…

mapkyca:
…lled elsewhere, it should probably be access restricted


mapkyca:
Interesting, this build failed on two of the mongo builds with a missing mongoclient.

I *suspect* this is a travis error, but will try and confirm.

mapkyca:
Build failures appear to be an intermittent problem, and my hunch is that it's something to do with the travis build image. Possibly related to this bug: travis-ci/travis-ci#4104

PR #1114, opened 2015-12-21
IndiePub: be our own micropub authorization_endpoint

kylewm:
Supports micropub for multi-user sites and doesn't rely on indieauth.com

Users are currently identified by the full path to their profile
/profile/username.  It would be nice in the future to allow
single-user sites to opt to use the bare domain as their ID instead.

TODO: Should we consider reusing/refreshing existing access tokens 
to avoid saving every generated token forever?


benwerd:
Super-cool!

Issue #918, opened 2015-06-01
Chicken and egg on registration of hooks


mapkyca:
This may be an edgecase, but it is not possible to have a login success/fail event triggered to guard APILogin via a plugin, because plugins aren't booted at this point.

But, I believe there was a prior issue which meant that API users needed to be logged in before their events were registered.

Perhaps there's a case for a pre-boot event...?

mapkyca:
Bumping this, this is far from an edge case.

Not having event hooks registered basically prevents third party authentication plugins like mapkyca/KnownOAuth2 from functioning. 

(Discovered while debugging mapkyca/KnownKnown#1, linked for reference)

kylewm:
I think this is closely related to a problem with IndiePub too. By the time it handles the request and decides the user's indieauth token is valid and logs them in, everybody else has already set themselves up in the "logged out" state.

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/918#issuecomment-168234951" class="u-in-reply-to"></a>I fixed it for the api by triggering the API login code after the instantiation of Plugins(). I suspect it might just be that we need to re-run the session boot at the end of the Plugin boot.

Little bit of a fudge, but might work for this too...

kylewm:
Any chance that means that plugins can register a hook on 'user/auth' and that hook will be invoked before every authed request? Specifically, I would like to do [this block](https://github.com/idno/Twitter/blob/master/Main.php#L35) in response to authorization instead of during init.

Or is your change more specific to API usage?

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/918#issuecomment-168238530" class="u-in-reply-to"></a>My patch fixes it for API only I'm afraid (it was the lowest impact change as I didn't have much time to test everything), however something similar could be adopted - reset the session user as a result of an auth trigger after plugin boot...

kylewm:
Thanks that makes sense! OK, will need to give it some more thought.

benwerd:
I'd hoped to get out a release today, but I'm in urgent care instead
(booooo). Which means we have time to get in a fix for this before the 0.9
release. It would be really nice to get this properly working.

Rearranging the auth code to run after a boot event seems very reasonable.

On Thursday, December 31, 2015, Kyle Mahan <notifications@github.com> wrote:

> Thanks that makes sense! OK, will need to give it some more thought.
>
> —
[...]

PR #1123, opened 2015-12-31
Refs #918 - Attempt API login after plugins have been booted.

mapkyca:


benwerd:
Hmm. Any clue on what's conflicting here?

kylewm:
It's definitely conflicting with #1125 which should be accomplishing the same thing (@mapkyca's PR moved the auth for API requests before plugins are created; I moved the auth for *all* requests). I think this one can be closed if Marcus agrees.

mapkyca:
Probably.

Close if to now, I'll confirm the API when I get a chance...

Issue #519, opened 2014-10-11
Quill can't find syndication links


jonpincus:
When posting to a known site, Quill can't find any syndication links.  The message:

No syndication targets were found on your site. You can provide a list of supported syndication targets that will appear as checkboxes here.

According to https://quill.p3k.io/docs#syndication :

Your Micropub endpoint will need to respond to a GET request containing a query string of q=syndicate-to. This request will be made with the access token that was generated for this app, so you can choose which syndication targets you want to allow this app to use.

kylewm:
I'm interested in trying to fix this. There's [some code](https://github.com/idno/Known/commit/6405fbbb71ecaa255b7a6b2934cbc9f470500abb#L28) that looks like it should get a list of endpoints but it's always empty for me (self-hosted or *.withknown.com). Will keep digging.

kylewm:
Best guess: syndication plugins register services (`Syndication::registerServiceAccount`) based on the current user during initialization, but for micropub requests there is no current user.

Micropub/Endpoint calls `Session::refreshSessionUser` to log the user based on an access token, but that happens way after all the plugins have been initialized.

I can think of a few approaches, all pretty bad:

1. Log a user in via micropub access token earlier in the initialization procedure, before any plugins are initialized. Would require micropub code outside of a plugin. :-1:
2. Syndication plugins all listen for user/auth event and re-registerServiceAccounts in response. Requires a change to every syndication plugin :unamused:
3. Micropub dumps and reloads plugins. Doesn't seem feasible; in particular I don't think there is a mechanism for destructing a plugin and deregistering its event hooks and such :-1:

kylewm:
hmm, and Convoy doesn't register service accounts locally at all. maybe that's a clue...

benwerd:
That's definitely a clue. But even with local accounts, the micropub implementation probably won't support syndication, in part due to a variable conflict. I think we should be able to expose accounts from Convoy though.

kylewm:
> the micropub implementation probably won't support syndication

like it won't right now, or it won't ever? If the latter, I'm a little confused because it seems like syndication should be possible using the same hook the native UI triggers.

kylewm:
@benwerd do you mean it won't work right now, or it won't/can't work ever? I'm curious what you mean about a variable conflict, I'm not familiar with that term.

[...]

PR #1125, opened 2016-01-02
sort out order of plugin registration and user auth

kylewm:
- clarify that "login" happens once when the user enters their
  username/password, whereas "auth" happens per-request
- plugins should expect to register hooks before user auth happens --
  cannot depend on the user being set during init.
- plugins should handle event "user/auth/request" to provide own
  authentication mechanism
- plugins should handle event "user/auth/success" to react to the user
  being authed
- updated IndiePub plugin to use the pluggable auth mechanism

Intends to address #519 and #918, and supersedes @mapkyca's PR #1123

### DIFF
--- a/IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
+++ b/IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
@@ -11,152 +11,115 @@
 
             function get($params = array())
             {
-
-                $headers = $this->getallheaders();
-                if (!empty($headers['Authorization'])) {
-                    $token = $headers['Authorization'];
-                    $token = trim(str_replace('Bearer', '', $token));
-                } else if ($token = $this->getInput('access_token')) {
-                    $token = trim($token);
-                }
-
-                if ($this->validateToken($token)) {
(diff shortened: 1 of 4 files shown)


IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
```diff
@@ -11,152 +11,115 @@ class Endpoint extends \Idno\Common\Page
...
-                if ($this->validateToken($token)) {
-                    if ($query = trim($this->getInput('q'))) {
-                        switch ($query) {
-                            case 'syndicate-to':
-                                echo json_encode([
-                                    'syndicate-to'          => \Idno\Core\Idno::site()->syndication()->getServiceAccountStrings(),
-                                    'syndicate-to-expanded' => \Idno\Core\Idno::site()->syndication()->getServiceAccountData()
-                                ], JSON_PRETTY_PRINT);
-                                break;
+                $this->gatekeeper();
```
kylewm: is this enough to prevent unauthorized acces?

benwerd:
Thank you!

PR #1129, opened 2016-01-06
IndiePub: fetch syndication targets from Convoy

kylewm:
Support ?q=syndicate-to queries on Convoy and hosted.

I couldn't find a particularly elegant way to do this, so I went with targeted/low-churn instead. IndiePub makes a request directly to the hub for the syndication targets. It gets back the HTML buttons and parses out the values from there. 

I'd also considered adding some deferrals in `Idno\Core\Syndication`, so that it could fetch the list of accounts the first time it's needed. I ended up not pursuing this because it would block rendering of the entire edit widget until you can connect to Convoy.


kylewm:
I think this is the last piece for #519

benwerd:
This. Is. Amazing. Thank you @kylewm!

PR #1131, opened 2016-01-10
Autosave: fixes to work with TinyMCE

kylewm:
- Use unique ID as selector when saving body
- Restore autosaved fields in getContent instead of the template.
- DOMXPath constructor causes a silent fatal error instead of throwing
  an exception; add null-safety for it


benwerd:
Amazing.

kylewm:
Interesting, richtext.tpl.php is used by more than just Text types. At least Photos and Checkins. I'll need to work on this a little more.

PR #1136, opened 2016-01-13
Call db->optimize at most once a day

kylewm:
Loading /admin/ takes a while because of a call to db->optimize().
This PR throttles it so that it will only call optimize if it's been
more than 24 hours since the last call. I'm open to other ideas too
(including leaving it alone)


physcocode:
+1 i was actually thinking about this

PR #1143, opened 2016-01-18
Collection validation

### LINKED ISSUE #1142: Whitelist (or otherwise validate) $collection in queries
Database queries sort and select via $collection (e.g. $collection = 'entities') in code.

This is used to construct a sql query dynamically, and although the parameters of this query are parameterised/prepared using PDO, this string is directly inserted.

This is bad, as a malicious bit of code/plugin/screwup could inject SQL into the query. Not very likely, but nonetheless is possible.

Since collections are known ahead of time and not very likely to change, we should validate this value, and perhaps do a whitelist against some "get collections" function.

mapkyca:
Closes #1142

### DIFF
--- a/Idno/Core/DataConcierge.php
+++ b/Idno/Core/DataConcierge.php
@@ -424,6 +424,18 @@
 
                 return array('$or' => array(array('body' => $regexObj), array('title' => $regexObj), array('tags' => $regexObj), array('description' => $regexObj)));
             }
+            
+            /**
+             * Internal function which returns an array of valid collections.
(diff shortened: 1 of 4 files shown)


kylewm:
Ohh good looking out.

Just to play devil's advocate, is there any possibility a plugin would add its own collection? I'm wondering if it might be preferable to check that the collection matches `[a-zA-Z0-9_]+`

mapkyca:
Hmm.. that's probably a good idea, aye

PR #1156, opened 2016-01-27
indiepub: trigger events before and after micropub save

kylewm:
add some hooks before and after publishing via Micropub.

needed these for https://github.com/kylewm/KnownInstagramShim


kylewm:
Rebased!

Issue #831, opened 2015-03-30
Change how API key validation works


mapkyca:
Problem:

API keys are passwords, and are stored cleartext in the database, and are generated automatically for a user. 

When an attacker gets read access to the database (or sniffs the traffic between the load balancer, web server or database server which is all unencrypted) they'll get access to all user accounts. This basically makes the effort gone to hashing passwords entirely pointless.

Solution:

Due to the way that the API works, it's pretty hard to simply hash the api key. Certainly, the api key should not be generated automatically, it should be explicitly generated by the user (to limit scope of a compromise).

That the signature is formed from hashing together the uri with the key means we can't simply store the key hashed and do a lookup, so I'm unclear at this point how the current HMAC signing authentication can be preserved. I'm starting to think we might be getting into ZKP territory, but I'll have a think.

Anyway, discuss.
##

benwerd:
This is one of those situations where I think it's helpful to look at other
peoples' API best practices. What are Twitter et al doing?

(Answer: OAuth and regenerating keys. Maybe there's something in there we
can pick up and use.)


On Mon, Mar 30, 2015 at 6:16 AM, Marcus Povey <notifications@github.com>
wrote:

> Problem:
>
> API keys are passwords, and are stored cleartext in the database, and are
> generated automatically for a user.
>
> When an attacker gets read access to the database (or sniffs the traffic
> between the load balancer, web server or database server which is all
> unencrypted) they'll get access to all user accounts. This basically makes
> the effort gone to hashing passwords entirely pointless.
>
> Solution:
>
> Due to the way that the API works, it's pretty hard to simply hash the api
> key. Certainly, the api key should not be generated automatically, it
> should be explicitly generated by the user (to limit scope of a compromise).
>
> That the signature is formed from hashing together the uri with the key
> means we can't simply store the key hashed and do a lookup, so I'm unclear
> at this point how the current HMAC signing authentication can be preserved.
> I'm starting to think we might be getting into ZKP territory, but I'll have
> a think.
>
> Anyway, discuss.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/idno/Known/issues/831>.
>



-- 
Ben Werdmuller
<http://goog_1933028737>
benwerd.com | werd.io

+1 (312) 488-9373

mapkyca:
Well, there's my oauth server plugin, which I think we talked about bringing in to core in the past.

[...]

PR #1164, opened 2016-01-29
Refs #831: Admins can't be admins when connecting via the API.

mapkyca:


benwerd:
For now - but once it's a more OAuth-like API auth mechanism I want to bring it back.

mapkyca:
Check my oauth plugin - perhaps some dusting off then bring into core?

mapkyca:
Check my oauth plugin - perhaps some dusting off then bring into core?

mapkyca:
Check my oauth plugin - perhaps some dusting off then bring into core?

PR #1168, opened 2016-01-29
ignore addtl clicks on contentCreateButton

### LINKED ISSUE #1167: Can open multiple editors on a slow connection
On a slow connection, I clicked "Status" to post a new status, and then clicked it again a few seconds later. The first editor popped up, I started typing, and then the second editor popped up erasing the first editor.

Probably need a flag to lock out the create buttons while an editor is loading/visible.

kylewm:
Use a flag to avoid the situation where you can click a content create
button multiple times. Opening a second editor on top of the first
editor can erase content (if you're on a slow enough connection that
it is noticeable).

Fixes #1167

### DIFF
--- a/js/default.js
+++ b/js/default.js
@@ -21,7 +21,15 @@ function bindControls() {
     $('.ignore-this').hide();
 }
 
+var isCreateFormVisible = false;
+
 function contentCreateForm(plugin) {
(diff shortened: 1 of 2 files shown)


benwerd:
You, sir, are on a roll. Thank you so much for this and everything you've committed.

Issue #1064, opened 2015-09-23
Convert Known dependencies to optionally use Composer


benwerd:
Because it's 2015.
##

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/1064" class="u-in-reply-to"></a>Y'mean instead of using github modules or including native?

Not entirely sure how you could do it optionally and not introduce an extra step, and if you do that you might as well use native modules rather than make everyone wget rootkit.php ...

kylewm:
I'd still love to see this, and before 1.0 is a good time IMHO. 

Rather than making it optional, I'd vote to adopt it and remove Symfony's UniversalClassLoader altogether. You'd still package the `vendor/` folder for releases, so people downloading a zip wouldn't need to install or use composer. But it would add an extra step for folks checking out from git.

In addition to all the nice stuff Composer gives you, I think it would remove a barrier to contribution... 

I started messing with it here, and it was surprisingly easy to get up and running:

  https://github.com/idno/Known/compare/master...kylewm:composer

Other than AutoP.php, there were packages for everything I looked for, and I think I more or less found the correct versions.

benwerd:
Composer is the standard way to include PHP modules in a modern project, so, okay: let's take another look. :)

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/1064#issuecomment-177046710" class="u-in-reply-to"></a>My €0.02 is that including dependencies in the binary are preferable as it gives you everything ready to go and there's less chance of tampering. 

Git submodules are next preferable because since it's a tool you're likely to have installed and you can peg modules to a known release.

Composer in my view is convenient (although not that convenient because you still have to install composer), but it scares the shit out of me because literally the first thing they tell you to do is "Please download and run this unverified script directly off the internet, but trust us it'll be *fiiiiiiine*".

Or as I said in my previous comment, "wget rootkit.php"

physcocode:
+1 to @mapkyca  composer is always the last choice for me including binary in the dependencies is cool as it removes the extra step of composer update .I am not totally against composer but i like the way know is currently loaded independent of a any extra step that composer would add

kylewm:
[...]

PR #1174, opened 2016-02-01
use Composer

kylewm:
Still very open to discussion of whether we really want to do this at all, but I'm sending the PR now in case it furthers that discussion.

And obviously this would be for *after* 0.9 ships.

Ref #1064


benwerd:
Following subsequent comments from @kylewm, I'm closing this. Let me know if you think I should reopen.

kylewm:
I think that's the right thing to do for now :cry:

benwerd:
Boooooo. I'm sorry. :(

It sounds like the conflicts were mostly to do with plugins? I wonder if there's another way to manage that?

kylewm:
> It sounds like the conflicts were mostly to do with plugins?

Yep

Ideally I guess I'd want a way of vendoring libraries so that they were namespaced specifically to that plugin. Then there is no way `IdnoPlugins\Reactions\Mf2\Parser` would conflict with `IdnoPlugins\IndieSyndicate\Mf2\Parser`. But that doesn't seem very realistic, since you'd have to edit your code and the code of the libraries that reference other libraries.

If we wanted to go all-in on Composer and use it to install the actual plugins themselves, then I guess it could resolve the conflicts. But that would mean nobody could install plugins without cli access and composer, and that's a non-starter.

PR #1178, opened 2016-02-02
Removes inline javascript

mapkyca:
* Removes inline javascript
* Moves embedding code to its own file
* Introduces a known javascript environment object
* Avoids a potential (although hard to exploit) XSS issue in display URLS


kylewm:
Woo, this is great! Can `annotateContent` move inside shell.js too? I don't *think* there's a reason it needs to be global?

mapkyca:
Probably doesn't need to be; updated....

physcocode:
I was thinking out loud about require.js !

mapkyca:
Resolved conflicts

mapkyca:
Conflicts resolved, back ported divergent changes (isLoggedIn()) and recommitted. 

Bumping this, as we should probably merge soon if we're going to.

mapkyca:
@benwerd ?

kylewm:
in the words of Joe Armstrong "This is good shit" :shipit:

PR #1189, opened 2016-02-03
No forwarding

mapkyca:
Pages now return correct error code (as opposed to forwarding and losing the response code).

### DIFF
--- a/Idno/Common/Page.php
+++ b/Idno/Common/Page.php
@@ -540,14 +540,14 @@
             function gatekeeper()
             {
                 if (!\Idno\Core\Idno::site()->session()->isLoggedIn()) {
-                    $this->setResponse(403);
+                    $this->deniedContent();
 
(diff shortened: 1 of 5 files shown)


benwerd:
Uh ... thinking about the user experience of this. Does this display the form on deniedContent when you're logged out? That would be neat.

physcocode:
@benwerd i think this actually does show the login form on denied content

mapkyca:
At the moment it doesn't.

I did originally include the login page on deniedContent, which worked, but 1) looked horrible ;) and 2) there's a case for having one page for login as a central point to override.

I was thinking about this on the way home from training, and I think maybe a good approach would be a "soft" redirect on deniedContent (window.location or meta refresh). That would preserve outward behaviour for users, but would mean you *would* get the correct HTTP error code back.

benwerd:
That makes sense. User experience has to trump API considerations, always - but this would satisfy both.

Issue #206, opened 2014-03-29
Retroactive posse


mapkyca:
Be able to posse already posted items in edit

benwerd:
Bumping. Retroactive POSSE and de-POSSE both.

mapkyca:
Bumping again, since it's been brought up in discussion more than once recently...

benwerd:
Yeah. Would be super-cool. Haven't had time, and the UX is important, but it's massively necessary.

mapkyca:
Bumping this again. Might try and implement it shortly...

benwerd:
Nice. I definitely have some thoughts on this. We do have some of the components to make this work.

kylewm:
I'm thinking through the UX a little, and it's definitely tricky. In Red Wind, it's always worked like: you edit a post, toggle the targets you want to syndicate to, and click save to POSSE to them. Clicking "Edit" in order to syndicate has always felt a little off to me; like clicking "Start" to shutdown the system.

I played around a little with an alternative; a screen specifically for syndicating. You click the "Syndicate" link

![syndicate-link](https://cloud.githubusercontent.com/assets/950127/12282761/83ade84a-b955-11e5-8515-6d8210191069.png)

to bring up the syndication page which just has the item rendered and the syndication toggles at the bottom

![syndicate-form](https://cloud.githubusercontent.com/assets/950127/12282762/85be9260-b955-11e5-896d-e684eeb59bac.png)

I dunno, it's functional but definitely feels janky in its own way.

And oh man de-POSSEing would be amazing (I see you feature creeping!), and definitely complicates things. Accidentally deleting a POSSE copy could be a disaster (I'm imagining like deleting a Medium post and losing a permalink that's been shared around). That *doesn't* really feel like something you should be able to do from the edit screen (?)

benwerd:
I'd say if you can remove the POSSE from the edit screen, it hits you with a bunch of confirmation errors. Not something you can accidentally click. Having the mechanism to remove POSSE also allows you to do things like effective snapchat: post a status update on Twitter for 30 seconds and then pull it, automatically. Suddenly content only needs to show up on a social network for as long as you want it to.

kylewm:
> 2) if ->posse contains $service we set disabled

There's a bit of an impedence mismatch now that each service has account as part of its identifier too

[...]

PR #1190, opened 2016-02-03
Retroactive posse

matzoman:
#206 

Adding the possibility to POSSE an entry in edit mode. The syndication toggles are disabled when the entry was POSSE'd already. (Disabling only works when the twitter/fb/etc. plugin adds the `Entity::setPosseLinks` parameter 'account_id'.)

major changes:
- a new `Entity::publish()` function all plugins must use for the button "Publish".
- the function `Template::drawSyndication()` has a new parameter (the return value of `Entity::getPosseLinks()`)

I changed all plugins to use publish() and added the new parameter to all drawSyndication calls.


physcocode:
wow this is blowing my  mind :+1:

kylewm:
Insert that Tim and Eric mind blowing gif. Just checked this out on my install, and it works beautifully! Nicely done!

benwerd:
I'm really looking forward to trying this out! Amazing work!

On Wednesday, February 3, 2016, Kyle Mahan <notifications@github.com> wrote:

> Insert that Tim and Eric mind blowing gif. Just checked this out on my
> install, and it works beautifully! Nicely done!
>
> —
> <https://github.com/idno/Known/pull/1190#issuecomment-179351288>.
>


-- 
Ben Werdmuller
<http://goog_1933028737>
benwerd.com | werd.io

+1 (312) 488-9373

benwerd:
Awesome!

kylewm:
It's worth noting that now `Entity::saveDataFromInput` calls `Entity::publish` which calls `Entity::save`, which isn't super intuitive. Would it be worth renaming saveDataFromInput? `publishDataFromInput` doesn't make much sense; maybe `processFormInput` or something??

Granted renaming across plugins is a pain, but if we're going to do it might as well do it now.

matzoman:
Thanks for all the kind words! It is a really engaging atmosphere in this project..

danito:
:+1: 
Just my feature request: can we change the colour of an already syndicated service? When I hit edit now, I don't know if I already have pushed it to tw or FB (without checking first), is there an easy way to change the grey toggle to maybe black or orange where the posse link exists?

srcr:
Awesome feature, thank @matzoman +1

Issue #1042, opened 2015-09-11
Tags from Instagram broken


davidmead:
Seems the latest updates have "broken" tags being passed from Instagram. They are being put back into the title, instead of the caption.

<img width="778" alt="screen shot 2015-09-10 at 9 46 59 pm" src="https://cloud.githubusercontent.com/assets/1266930/9805291/86167716-5805-11e5-94fe-262270e490f4.png">

http://blog.davidjohnmead.com/2015/washed-rosie-campbowwow-vscocam-365daysofme-x6-dog

davidmead:
Added my code from https://github.com/davidmead/IndiePub/commit/04d66b8e160c26dbe5e71643db6dcc4ec1e4ea42, back into endpoint.php. Seems to have added back the `u-syndication` links from Instagram posts.

kylewm:
fixed by #1193!

PR #1193, opened 2016-02-04
upgrade micropub endpoint

matzoman:
Adding support for `'syndication'` and `'category[]'`:
- the field syndication gives us an URL to add a POSSE link for the new entry
- the array category[] is translated into hashtags in the content
- furthermore the hashtags from the `$name` are removed if there are categories delivered

(Also a solution to #1042 "Tags from Instagram broken")

The service [ownyourgram ](http://ownyourgram.com/) i.e. send these fields. With this patch it is possible to collect 'likes' & 'comments' via brid.gy

Right now the name of service is determined with parseUrl. maybe there is a better, but still generic way to do this ?!

---

An example on my blog: http://known.matthias.welling.io/2016/2-fruehstueck


kylewm:
This looks like a pretty decent general solution to me. 

Previously discussed the issue with syndication links from OwnYourGram here: https://groups.google.com/forum/#!topic/known-dev/QZviPC5w9oM ... and much later I wrote a plugin to jam the instagram identifiers in there when a syndication link from IG is detected https://github.com/kylewm/KnownInstagramShim

This has the disadvantage of being specific to Instagram but the advantage of matching the appearance of syndication links to other platforms:

screenshot from 2016-02-04 08-15-37

Also a note: unfortunately Known uses the same parameter ("syndication") to mean "platforms to syndicate to"

PR #1197, opened 2016-02-04
Reconcile DataConcierge::getAnyRecord with other DBs

### LINKED ISSUE #978: Mongo and Mysql data sources have inconsistent behaviour

kylewm:
Mongo's getAnyRecord always returned an array while
other implementations converted the array to an object
if possible.

Changed all DBs to return an array for getAnyRecord and
added DataConcierge::getAnyObject.

(Config was the only thing using this function
and this allows us to remove some special casing from its
init)

Fixes #978

### DIFF
--- a/Idno/Core/Config.php
+++ b/Idno/Core/Config.php
@@ -110,9 +110,8 @@
              */
             function load()
             {
-                if ($config = \Idno\Core\Idno::site()->db()->getAnyObject('config')) {
+                if ($config = \Idno\Core\Idno::site()->db()->getAnyRecord('config')) {
                     $this->default_config = false;
-                    $config = $config->getAttributes();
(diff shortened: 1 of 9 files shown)


kylewm:
Hmm, this is trickier than I thought. Going to close and go back to the drawing board.

Issue #1177, opened 2016-02-02
\Idno\Core\Logging should follow PSR-3 standards


physcocode:
with reference to #1133 would be great for supporting multiple logging engine

physcocode:
I was working on a PR for this just wanted to make sure are you interested in it @benwerd ?

any suggestions @mapkyca @kylewm ?

kylewm:
@physcocode nice! I'd started working on this here https://github.com/kylewm/Known/commit/b103d49dc1d1ff38811079ba66a7b4facca99917 ... that was dependent on the Composer stuff and at the time I was thinking we'd include Monolog in core. But maybe that can save you the trouble of changing all the log() calls yourself.

One bit that I think is particularly useful: https://github.com/kylewm/Known/commit/b103d49dc1d1ff38811079ba66a7b4facca99917#diff-e3f0ea8006f55c95d9bf4d5e689e74caR176 maintains backward compatibility with the existing `log($message, $level);` method by detecting this case and reversing the parameters. In a future version, I think we should add a warning that this order is deprecated, and in an even later version remove it.

physcocode:
@kylewm you made my day :) thats gonna save a lot of labour :+1:

physcocode:
well this boosted my approach ill send a pull request in few hours :)

Issue #1133, opened 2016-01-12
Consider using monolog (or equivalent) for Logging


kylewm:
Maybe it's my newness to the php ecosystem, but logging via `error_log` feels limited. `\Idno\Core\Logging` helps by adding some context. I wonder if we should consider replacing it with the PSR-3 compatible Monolog.

I think this would be nice for a few reasons:
- Seeing incomplete entries in log file: I'm not sure if this was because the request terminated before it finished writing the log or if it reached some size limit, but when I was debugging issues with the Flickr API, I routinely had log messages cut off in the middle.
- Log messages wouldn't have to go through nginx/apache. I'm not sure this is important, but I feel better knowing messages are being written directly to file and not being cut off by the reverse proxy.
- Variety of Handlers allow logging to files (including timestamped files), the javascript console, FirePHP plugin, or email. I email errors to myself in all my Flask projects, and it's great.

benwerd:
I think this is a great idea, tbh. Being able to leverage other projects like this rather than reinventing wheels sounds like the way to go. Will take a deeper look.

benwerd:
A good middle ground could be to establish \Idno\Core\Logging as an effective wrapper for Monolog, should we want to change it out in the future.

kylewm:
I'm having second thoughts about this. 

I'm not convinced that deferring to a logging framework is a good idea for every user ... it could lead to a situation where you have to check the Known log *and* the error log to track down issues, and it's already hard enough to explain to people where the error log might be.

For now I think it would be worthwhile to make `\Idno\Core\Logging` implement [`\Psr\Log\LoggerInterface`](https://github.com/php-fig/log/blob/master/Psr/Log/LoggerInterface.php), and to make Monolog an optional plugin (maybe the plugin's init would replace `\Idno\Core\Logging` with its own `\Monolog\Logger`). Would also give us some time to evaluate whether it's actually a good thing for core.

physcocode:
+1 having abstract classes to be extending function is the way i support third party integration in my software mixed up with Dependency injection

physcocode:
Dependency Injection is not a cup of tea for known but there should be abstract classes to extend some sort of core functionality like logging via a plugin . calling abstract classes instead of main class increases scope of changes with our rewriting the whole core . For example lets say in some far future known want to render classes via a futuristic tempting engine and not Bonita . This would be a lot of rewrite either in core of known or the tempting engine , but having abstract classes as a middleware solve this kind of thrones in development and extension

[...]

PR #1198, opened 2016-02-04
PSR-3 logging in known

physcocode:
As discussed in #1177 and #1133 this added following ability.
1) The default known logger follow PSR-3 standards
2) Any PSR-3 compatible logger can replace the known logger
3) Plugins can override default logger using \Idno\Core\Logging::setLogger() [As demanded in #1133 ].
4) \Idno\Core\Logging::Log() method still works for backward compatibility.


physcocode:
I have a question : If plugin set logger , anything logged before plugins are loaded will be logged by default logger won't it cause inconsistency ? @kylewm @benwerd  any thoughts ?

kylewm:
@physcocode thanks for making these changes! I left some inline comments, hope they are useful :)

I agree that there's a problem with registering a Logger via a plugin, i.e. all log messages before the plugin are loaded would go to the default logger. We might consider (not necessarily for this PR) adding some parameters to the .ini file so you could load and configure the proper logger during startup.

Idno/Core/KnownLogger.php
```diff
@@ -0,0 +1,161 @@
...
+         */
+        public function notice($message, array $context = array()) {
+            $this->log($message, 3, $context);
+        }
+
+        /**
+         * Writes critical message to error log
+         * @param string $message
+         * @param array $context
+
```
kylewm: nit: trailing whitespace :P

physcocode:
@kylewm The logger is kind of a wrapper class , any class implementing LoggerInterface can be Dependency Injected via setlogger() in this way KnownLogger can be kept very much separate any other logger like monologger can easily and totally replace Known Logger , If KnownLogger and Logger are merged no other logger class can totally replace it.

In a nutshell Logger class is somewhat like DataCongire . You should Read about DEPENDENCY INJECTION to know its advantage.

as @benwerd said `I think this is a great idea, tbh. Being able to leverage other projects like this rather than reinventing wheels sounds like the way to go.  ` dependency injection is the solution of this problem and Logger is the Proof of concept . Any other class can also be changed in the same way so that any third party tool can be integrated without changing single line of core in future .

physcocode:
Hold On ! I am revising the pull request solving problems @kylewm commented , and may break the backward compatibility to log() function due to the $level , $message swapping .  Too Bad PHP does not support function overloading , that was the only way of backward compatibility

physcocode:
for a bonus @kylewm i already made a plugin to implement monolog will surely make it available if this is merged in core

Idno/start.php
```diff
@@ -94,9 +94,15 @@ function &loader()
 
 // Bonita is being used for templating
     $known_loader->registerNamespace('Bonita', dirname(dirname(__FILE__)) . '/external/bonita/includes');
+//PSR Log Interface    
+    include dirname(dirname(__FILE__)) . '/external/Psr/Log/LoggerInterface.php';
+    include dirname(dirname(__FILE__)) . '/external/Psr/Log/AbstractLogger.php';
+    include dirname(dirname(__FILE__)) . '/external/Psr/Log/LogLevel.php';
+
```
kylewm: This should not be necessary - you should be able to add external/PSR with the loader's registerNamespace function instead.

Idno/Core/KnownLogger.php
```diff
@@ -0,0 +1,75 @@
...
+                    $level = "WARNING";
+                if ($level == 3)
+                    $level = "INFO";
+                if ($level == 4)
+                    $level = "DEBUG";
+                // Logging contexts
+                $context='';
+                if (!empty($context)) {
+                    $context = ' [' . implode(';', $context) . ']';
+                }
```
kylewm: this clears the context and then immediately checks if it's empty? I think there was some logic lost in the move from Idno\Core\Logger

Idno/start.php
physcocode: @kylewm If you see my previous commit i used the registerNamespace function but it did not work idk why , please do try and tell me if it worked for you

physcocode:
@kylewm thanks for the catch it has been fixed :)

kylewm:
> You should Read about DEPENDENCY INJECTION to know its advantage.

Thanks, I know about dependency injection. Notice that it is just as easy to call `Idno::site()->setLogger(...)` as it is to call `Idno::site()->logging()->setLogger(...)`

> break backward compatibility to log() function due to the $level , $message swapping . Too Bad PHP does not support function overloading , that was the only way of backward compatibility

I can see keeping \Idno\Core\Logging as a wrapper around KnownLogger *if* `\Idno\Core\Logging::log()` is made to support backward-compatibility with old-style `->log("message", LOGLEVEL_DEBUG)` calls. As suggested previously, you can check `is_int($message)` to tell if it's being called with the old signature.

Your KnownLogger::log implementation is still expecting to get $level as an integer; it should be expecting a string (`\Psr\Log\LogLevel`) instead. 

I still think it makes sense for \Idno\Core\Logging to implement `\Psr\Log\LoggerInterface`.

Also, it'd be nice to squash these commits into one or two atomic changes.

Idno/start.php
kylewm: If you clone https://github.com/php-fig/log into external/, then you should be able to call
```php
registerNamespace('Psr\Log', __DIR__.'/external/log');
```
I think the issue right now is that you have the Psr namespace right in external rather than a subdirectory like the other external packages

physcocode:
backward compatibility is not possible due to the swap . I'll keep the PR as it is for now :)

physcocode:
@kylewm I got your point i am sending new PR

PR #1201, opened 2016-02-05
PSR-3 logging in known

physcocode:
1) Set Logger to follow PSR standards
2) Plugin can replace logger using \Idno\Core\site()->setLogger()

Note : breaks backward compatibility
All classes are changed to work with PSR-3 logger


physcocode:
Solves #1177 and #1133 , improved version of #1198 :) 
special thanks to @kylewm for correcting all my inappropriate approach :)

PR #1207, opened 2016-02-06
Adds the ability to set a static page as the homepage.

smartboyathome:
I'm new to Known, and my day job is not PHP so this probably needs some cleanup.

Anyway, my primary motivation for this plugin was to make it easier to customize the homepage without too much work creating plugins and themes. I knew that I had the ability to create static pages, but none of them could be promoted to the homepage.

This adds the ability to set a page as your homepage by pressing the "home" icon on the pages admin page. In order for the homepage not to disappear completely, I needed to modify the "links" template in order to allow me to modify the default link. In addition, I added the ability to set a title on links created by "createLink" in order to condense the homepage toggle down to just an icon in order to save space.

I tried to make the former homepage discoverable, but I'm not in love with how this does that. Please, give me any and all suggestions you may have.


kylewm:
This sounds really cool. Do you have a public site you can share that has a static page as the homepage? Curious to see how it looks.

smartboyathome:
Indeed I do. Check out [my test site](https://strathis.com/). Eventually, I plan to upgrade this to my actual homepage running on top of Known, but for now I'm using it to test the waters.

Issue #170, opened 2014-01-16
Add in-Known notification pane


erinjo:
For a multi-user install like an intranet where everyone is on the same domain, what's the plan for handling mentions/responses between users?  Is there something in the works like an @-mention notification section or a recent comments section so that employees could easily see when another coworker posted a comment or left a mention on a status update?

benwerd:
I know this is on your tasklist, but bumping to indicate that it's still something we're interested in :)

benwerd:
Bumping this. Having internal notifications for, eg, new webmentions, new joined users, etc, would be really useful.

mapkyca:
Should this ticket be merged with #7 ?

benwerd:
Let's keep them linked, as technically one is notifications themselves, while the other is a native notification pane. Subtle difference, but still ;)

mapkyca:
Closing this, as I think this was addressed by @kylewm's work

PR #1218, opened 2016-02-11
Notifications

kylewm:
I'd love to get community feedback on this.

* store Notification entities in the datastore
* show a basic Notification panel for logged in users (under /account/notifications)
* show unread notifications count above the user name for logged in users.
* includes a change to Bonita to dump template vars into the local scope, so instead of referencing `$vars['object']` you can just say `$object`

Here's how it looks

notifications

Ref #7 and #170


benwerd:
WOW.

Wow wow wow.

This is so badly needed. I'm going to spend some solid time with this.

kylewm:
> I'm going to spend some solid time with this.

Awesome, thank you! 

I'm somewhat confident in the data model; I thought through the different notifications types I can imagine, and I believe it is flexible enough to accommodate them.

The actual UI/notifications panel itself, I'm much less sure about (unsurprisingly). For one thing, it uses a lot of vertical space for each notification, and may use Bootstrap panels too liberally.

PR #1221, opened 2016-02-13
Unify friending with share

mapkyca:
Adds friending workflow to the main bookmarklet


benwerd:
I'm holding off on this. I think there's a bigger conversation to be had here about the future of this functionality.

PR #1222, opened 2016-02-14
Closes #476: Better handle long names

### LINKED ISSUE #476: Better design handling around long names #enhancement
Explore adding extra padding, wrapping, or other options to prevent long names from bumping up against the posts.  
screen shot 2014-09-22 at 9 36 21 am

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4636561-better-design-handling-around-long-names-enhancement?utm_campaign=plugin&utm_content=tracker%2F300028&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F300028&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mapkyca:


css/default.css
```diff
@@ -143,6 +143,10 @@ img {
     line-height: 1.2em;
 }
 
+.idno-entry .h-card .p-name {
+    word-break: break-word;
```
kylewm: Is it possible this should be `word-wrap: break-word`? Based on MDN, I think so

https://developer.mozilla.org/en-US/docs/Web/CSS/word-wrap
https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

mapkyca: word-break is valid, but I think word-wrap is more compatible - added both.

PR #1228, opened 2016-02-14
Fix MySQL schema entry value to match update date

n-i-m-a:


kylewm:
great catch @n-i-m-a! I'm embarrassed that I missed it

PR #1231, opened 2016-02-15
Webservice::post should not form-encode params

kylewm:
leave the encoding to cURL. this is required in order to
send files (key=@filename) via POST for micropub

### DIFF
--- a/Idno/Core/Webservice.php
+++ b/Idno/Core/Webservice.php
@@ -68,19 +68,19 @@
                 switch (strtolower($verb)) {
                     case 'post':
                         curl_setopt($curl_handle, CURLOPT_POST, 1);
-                        curl_setopt($curl_handle, CURLOPT_POSTFIELDS, self::flattenArrayToQuery($params));
+                        curl_setopt($curl_handle, CURLOPT_POSTFIELDS, $params);
                         curl_setopt($curl_handle, CURLOPT_HTTPHEADER, array("Content-type: multipart/form-data"));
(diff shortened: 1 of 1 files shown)


mapkyca:
Hmm... I had a battle with this before when writing mapkyca/KnownKnown, and it definitely used to work. If it's stopped then that's a revision.

Worth checking against that, since it's a complete usage of the API.

kylewm:
Oh OK thanks for the heads up. It looks like the behavior changed in
14684b61c93b5c491bb543b4a733afa234d6f096

On Mon, Feb 15, 2016, at 12:56 AM, Marcus Povey wrote:
> Hmm... I had a battle with this before when writing
> mapkyca/KnownKnown, and it definitely used to work. If it's stopped
> then that's a revision.


> Worth checking against that, since it's a complete usage of the API.


> — Reply to this email directly or view it on GitHub[1].


>



Links:

  1. https://github.com/idno/Known/pull/1231#issuecomment-184120102

kylewm:
Yeah it looks like photos don't work Known-to-Known anymore. I got on the receiver:

```
PHP message: PHP Notice:  Undefined index: photo in /srv/www/kylewm.com/known/IdnoPlugins/Photo/Photo.php on line 153
```

and on the client

```
PHP message: Known (known.dev): INFO - API Returned code 400
PHP message: Known (known.dev): INFO - Details: array (
  'header' => 'HTTP/1.1 400 Bad Request
```

PR #1232, opened 2016-02-15
allow ContentTypes to override their default edit URL

kylewm:
fixes the remaining conflict between IndePlugins/Reactions/Like and
IdnoPlugins/Like/Like because Reactions can use any edit url it
prefers now (currently using /indielike/edit instead of /like/edit,
which is admittedly a little bit puerile, "my likes are indier than
yours")

### DIFF
--- a/Idno/Common/Entity.php
+++ b/Idno/Common/Entity.php
@@ -1632,7 +1632,7 @@
                     }
                 }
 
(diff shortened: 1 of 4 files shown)


kylewm:
any idea why Travis stopped testing PR's?

kylewm:
> any idea why Travis stopped testing PR's?

mm, they were just disabled in the Travis config. I turned them back on, but it doesn't look like it's retroactive.

PR #1236, opened 2016-02-15
Custom connection params

### LINKED ISSUE #1235: Pass database parameters as settings not from config
Settings for databases should be passed to an object's constructor, not magically pulled from $config inside the init.

This would allow for more than one database connection to exist.

Refs #1234

### LINKED ISSUE #1237: Abstract Dataconciege
DataConciege is actually Mongo, rather than a suitable parent for other data types. We should abstract this and then create a specific Mongo data object.

mapkyca:
Allows multiple instances of a database connection to be created, with their own connection constructors.

Begins the process of migrating mongo code into its own class, and abstracting DataConciege.

Closes #1235 
Closes #1237

### DIFF
--- a/Idno/Data/Mongo.php
+++ b/Idno/Data/Mongo.php
@@ -22,6 +22,12 @@ namespace Idno\Data {
         private $dbuser;
         private $dbpass;
 
+        /**
+         * Escape sequences for sanitizing fields that will be stored in Mongo.
+         * Note that % must be first so that it doesn't double-escape previous sequences
+         */
+        private static $ESCAPE_SEQUENCES = ['%' => '%25', '$' => '%24', '.' => '%2E'];
+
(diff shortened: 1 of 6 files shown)


mapkyca:
(love unit tests)

Idno/Data/Mongo.php
```diff
@@ -0,0 +1,407 @@
...
+ * 
+ * @package idno
+ * @subpackage data
+ */
+
+namespace Idno\Data {
+
+    use Idno\Core\Idno;
+
+    class Mongo extends \Idno\Core\DataConcierge {
```
physcocode: extends or implements DataConcierge?

kylewm: are you asking if this should be implements instead of extends? DataConcierge is still a class, not an interface, so extends is correct.

PR #1240, opened 2016-02-15
Refs #1238: Statement of intent - remove PHP 5.4 from unit tests

mapkyca:

PR #1243, opened 2016-02-16
Sunset PHP 5.4 support

mapkyca:
Contains patch from #1240, and modifies the docs accordingly.


mapkyca:
I think there's a bit of communication to manage here to give people sufficient warning, but 5.4 has been EOL for a year now, so really we should be dropping official support for the next release (obvs. I can't make an official statement on the matter).

Not that there are currently any breaking changes with 5.4 that I'm aware of, but it should be made clear to folk that 5.4 has gone the way of Windows XP and they need to migrate to something newer.

My personal feeling is that we should really be stating 5.6+, since 5.5 will be EOL in July, but one step at a time ;)

benwerd:
I addressed this on the mailing list - most hosts are still on PHP 5.4,
including many shared hosts. We can't sunset it yet. (Actually most hosts
are on 5.3, but come on.)

It isn't our job to take care of peoples' server security for them, and in
this case doing so would significantly reduce the number of people who can
use Known.

For reference: http://w3techs.com/technologies/details/pl-php/5/all


On Tuesday, February 16, 2016, Marcus Povey <notifications@github.com>
wrote:

> I think there's a bit of communication to manage here to give people
> sufficient warning, but 5.4 has been EOL for a year now, so really we
> should be dropping official support for the next release (obvs. I can't
> make an official statement on the matter).
>
> Not that there are currently any breaking changes with 5.4 that I'm aware
> of, but it should be made clear to folk that 5.4 has gone the way of
> Windows XP and they need to migrate to something newer.
>
> My personal feeling is that we should really be stating 5.6+, since 5.5
> will be EOL in July, but one step at a time ;)
>
> —
> <https://github.com/idno/Known/pull/1243#issuecomment-184635909>.
>


-- 
Ben Werdmuller
<http://goog_1933028737>
benwerd.com | werd.io

+1 (312) 488-9373

mapkyca:
Agreed, and I figured that'd be your answer (I didn't see your mailing list post, sorry). 

I agree it's not our job to look after other's server security, I'm more concerned with the technical debt of maintaining a requirement to support legacy language versions into the future. I admit that 5.4+ is less of a headache than keeping 5.3 support, but there are newer features and language constructs that are handy in 5.5+, and shimming for their 5.4 support is tricky.

Additionally, none of us actively develops in 5.4, which means we're increasingly going to get issues like the Photos plugin crash, which unit tests can only do so much to detect.

My reasoning for this patch is that 5.4 has been dead for a while, regardless of what the hosting stats say, and my thinking is that if we state we support later versions then that gives a little wiggle room when those things crop up with our code, or third party plugins (who definitely won't be as rigorous with unit tests as we are). 

Changing the docs and travis.yaml won't magically break 5.4 support, so folk are free to try it, but these installs are very much YMMV, and while we will try and keep things working, we can't guarantee it.

But obvs, if the decision is made, the decision is made... but we're going to have to end support for it at some point. A soft removal now followed by a hard removal once the stats drop below X seems reasonable.

mapkyca:
Ahh... I see the conversation, and I remember it now.

I still think my point stands, we've got to EOL support at some point. If we do this softly now (not officially, but making YMMV best efforts to support) then by the time there's a breaking change the install base should be small (given that 5.4 -> 5.6 upgrades are painless when compared to 5.3->5.4 which was a blocker for a lot of people).

mapkyca:
Conflicts resolved

mapkyca:
Set NRN for this, but with the full intention to revisit it Real Soon, once we have some data (e.g. by #1265)

PR #1244, opened 2016-02-16
bugfix for importing rss feeds

matzoman:
changing from `Entity::save()` to `Entity::publish()`

(I overlooked this call to `save()` when changing the behaviour for "retroactive posse" #206 )


benwerd:
Great catch. Thank you!

PR #1249, opened 2016-02-17
Micropub Endpoint creates checkin if mp-type == checkin

### LINKED ISSUE #1227: micropub checkins not working
just tested http://ownyourcheckin.wirres.net/ which peso your checkin from 4sq and publish it via micropub back to your site. The status is coming to my known site but not as checkin.
See https://nxd4n.nixekinder.be/2016/checkin-foursquare-4sq-1
This is what is posted:

```
[published] => 2016-02-14T20:04:57+01:00
[location] => geo:50.807987237976,4.313364561688
[place_name] => Georgiana Island
[place_url] => https://foursquare.com/venue/555b4630498eec90b9af983b?ref=YOFY2SHCUPCRRJUBFNI2D1SU3IKOFBHOGSX3ZN5CDEARGODM
[category] => Array
(
[] => checkin
[] => foursquare
[] => 4sq
)

[content] =>
[syndication] => https://foursquare.com/forward/checkin/56c0cfd9498e254d8a1ecb77
```` 
The checkin category is ignored or probably nothing checks against if it is possible to publish a checkin entity if it is enabled.

danito:
works with  @diplix http://ownyourcheckin.wirres.net
resolves #1227

### DIFF
--- a/IdnoPlugins/Checkin/ContentType.php
+++ b/IdnoPlugins/Checkin/ContentType.php
@@ -7,6 +7,7 @@
             public $title = 'Location';
             public $category_title = 'Locations';
             public $entity_class = 'IdnoPlugins\\Checkin\\Checkin';
+            public $indieWebContentType = array('checkin');
 
         }
 
--- a/IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
+++ b/IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
@@ -72,6 +72,7 @@
 
                 // Get details
                 $type        = $this->getInput('h', 'entry');
+                $type        = $this->getInput('h');
                 $content     = $this->getInput('content');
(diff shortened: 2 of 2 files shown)


danito:
I hope I got it right this time. (OT how can I get rid off those unnecessary commits when PR?)

kylewm:
Looks good to me! 

> how can I get rid off those unnecessary commits when PR?

`git rebase -i origin/master` is useful here -- it gives you the option to squash commits with their predecessor, so you can just change all but the first "pick" to "squash". if you've already pushed the branch, then you'll need to force push since this rewrites history.

but really if it's just a few commits, it's usually not necessary to squash them. i only get picky when there are a ton of commits, or they make and then undo unrelated changes (because it makes the version history confusing)

also: I'd recommend working in a branch `git checkout -b <branchname>` when you're putting together a pull request.

PR #1258, opened 2016-02-17
Make webservice calls statically

mapkyca:
Calling static methods non-statically causes notice error in PHP7


mapkyca:
Merging in master to get rid of the failing message

PR #1261, opened 2016-02-17
adding missing field and escaping not used variables

danito:
if entity is not a checkin an error is logged as specific variables aren't used.


benwerd:
Thank you!

IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
```diff
@@ -104,6 +104,7 @@ function post()
                     }
                       if ($type == 'checkin')  {
```
kylewm: is this working for you? we're inside the check that `$type == 'entry'`, so I would expect it to be impossible for it to be 'checkin' here

Issue #1149, opened 2016-01-21
Remove inline Javascript


mapkyca:
We need to stop putting javascript in the template as it makes it harder to maintain, interferes with CSP and if it's dynamically produced is all to easy to end up with DOM based XSS attacks.

Move all this gumph into its own files.

benwerd:
+1,000.

kylewm:
Semi-related to #914 (Consider using require.js)

physcocode:
Aint this covered under #1152 ?

PR #1268, opened 2016-02-18
Ref #1149: move checkins javascript to its own static js file

kylewm:
Refactor and move checkins javascript to checkin.js. Instead of checking object->id, this detects new vs. edit by checking whether lat and long have been set already.

Also fix a bug I introduced in the .gitignore file that was ignoring too much of IdnoPlugins.


mapkyca:
Nice. 

Refs #1178

PR #1281, opened 2016-02-21
Supports filtering of annotations

mapkyca:


mapkyca:
Check mapkyca/KnownAkismet :)

PR #1283, opened 2016-02-21
pull common SQL functions to AbstractSQL

kylewm:
- refactor to eliminate duplicate code between Postgres, Mysql, &
  Sqlite implementations
- change logic of build_where_query slightly to support
  `[$not]` and `[$not][$in]` (previously it combined the two cases).
- also check for `$key == '$or'` rather than `!empty($value['$or'])`,
  which I don't think made sense.

### DIFF
--- a/Idno/Core/DataConcierge.php
+++ b/Idno/Core/DataConcierge.php
@@ -16,17 +16,6 @@
         {
 
             protected $client;
-            protected $database;
-            
-
-            /**
-             * Returns an instance of the database reference variable
-             * @return mixed
-             */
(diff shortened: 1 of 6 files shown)


physcocode:
If we are dropping support for mongo can't DataConcierge itself be the AbstractSQL ? why two level of heridity ?

benwerd:
@physcocode I support this method - this way we can still support non-SQL databases in the future.

physcocode:
+1 @benwerd well i like reddis by the way :) i will someday try reddis with known :)

Issue #1257, opened 2016-02-17
fa fa-foursquare missing on MicroPub checkins from 4sq


danito:
When I checkin in 4sq and Micropub it, it gets a default fa-share icon and not the fa-foursquare icon.

kylewm:
Yeah this is a known issue with posts made by micropub. I wrote https://github.com/kylewm/KnownInstagramShim to wedge the instagram icon and username into posts coming from OwnYourGram. Foursquare is a bit different since Known already has a plugin for it, and may already have a syndication account registered where it can get the username and info from...

You could put some similar code to this https://github.com/kylewm/KnownInstagramShim/blob/master/Main.php#L22 in the Foursquare plugin. Or if you just wanted to add it to the InstagramShim plugin, that would be fine with me too as long as it doesn't break existing functionality.

danito:
will have a look. 
What I don't get is why [templates/default/content/syndication/links.tpl.php#L20](https://github.com/idno/Known/blob/master/templates/default/content/syndication/links.tpl.php#L20)
` foreach($posse_links as $element) {
                    $this->username = isset($element['account_id']) ? $element['account_id'] : false;
                    $human_icon = $this->draw('content/syndication/icon/' . $service);
                    if (empty($human_icon)) {
                        $human_icon = $this->draw('content/syndication/icon/generic');
                    }
                    echo '<a href="' . $element['url'] . '" rel="syndication" class="u-syndication ' . $service . '">' . $human_icon . ' ' . $element['identifier'] . '</a> ';
                }`

doesn't catch  the icon if I have installed the Foursquare plugin.

kylewm:
It's because the indiepub hack to use the domain name instead of the service name gives you service="foursquare.com" instead of service="foursquare"

danito:
why not add 
`$posse_service = str_replace('.com', '', $posse_service);` after

[IndiePub/Pages/MicroPub/Endpoint.php#L157](https://github.com/idno/Known/blob/master/IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php#L157)
````php
                        if (!empty($posse_link)) {
                            $posse_service = parse_url($posse_link, PHP_URL_HOST);
                            $entity->setPosseLink($posse_service, $posse_link, '', '');
                        }
````

if this to differentiate with the syndication service (pesso)?

kylewm:
This is fixed now, ya?

danito:
Closed by #1285

PR #1285, opened 2016-02-21
MicroPub Endpoint accepts Photo file from checkin

danito:
- gets $_FILE['photo'] and adds it to the checking
- multiple words categories changed to #MultipleWords 
- fixes $content override if $content was array.


danito:
Fixes #1289

danito:
Fixes #1257

Issue #1299, opened 2016-02-24
Checkin leaflet map : global variable map/marker


danito:
## While trying to do this:
editing the Foursquare plugin to address #1290 
## I encountered this error:
the leaflet map is not accessible for plugins to modify as the map is defined in a private function
## Some other notes:
If nothing speaks against it, make the map and the marker variable global in [IdnoPlugins/Checkin/checkin.js](https://github.com/idno/Known/blob/master/IdnoPlugins/Checkin/checkin.js)

physcocode:
this is fixed in #1300 right?

danito:
Yes

Issue #1290, opened 2016-02-22
Location & syndication to FourSquare : provide suggestions for place_name


danito:
## While trying to do this:

Creating a location and syndicate it to Foursquare 
## I encountered this error:

The provided placename is a street number of the found address or a location name. If this name doesn't correspond to the exact same 4sq name, your checkin is completely wrong.
## Some other notes:

I would provide a list of places to suggest where to checkin if the syndicate to 4sq is toggled on, based on the the LatLong values and not the location name to be sure to get only 4sq places.
Maybe change the placename field to a dropdown with the returned search results with a warning "please chose one of the suggested 4sq places or disable 4sq ".

danito:
I started to address this in a new [branch](https://github.com/danito/Foursquare/tree/4sqVenues).
Any comments are welcome.

benwerd:
Super super super cool!

On Wed, Feb 24, 2016 at 6:13 AM, Daniel nix <notifications@github.com>
wrote:

> I started to address this in a new branch
> <https://github.com/danito/Foursquare/tree/4sqVenues>.
> Any comments are welcome.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/idno/Known/issues/1290#issuecomment-188271899>.
>



-- 
Ben Werdmuller
<http://goog_1933028737>
benwerd.com | werd.io

+1 (312) 488-9373

mapkyca:
Closing, is handled by a (yet unmerged) PR on the Foursquare plugin.

PR #1300, opened 2016-02-24
Update checkin.js

danito:
copy map & marker to global variables to be able to use them outside this fct
## Here's why I did it:
to address #1290 and #1299


danito:
or is `var this.MyVariable` a better way to declare a global ?

Issue #235, opened 2014-04-24
Move get/set Input methods


mapkyca:
Because you quite often need to call these _before_ there's a current page...

benwerd:
Couldn't this be achieved with an event in getContent / postContent?

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/235" class="u-in-reply-to"></a>I don't believe so. There have been times where I've needed access to these in plugin init hooks, and this is before there's a current page defined. I've had to fall back to using the unfiltered variables directly, which is less than idea.

So, current page either needs to be instantiated before plugin hooks are registered, or we need to move input functions somewhere else.

benwerd:
It does make some sense to be able to access input variables in a way that's untethered from the current page.

PR #1303, opened 2016-02-25
Assets array needs to be static to avoid assets being set on wrong pa…

mapkyca:

Set assets array to static

## Here's why I did it:

Assets set in plugins were being lost.

### DIFF
--- a/Idno/Common/Page.php
+++ b/Idno/Common/Page.php
@@ -49,7 +49,7 @@
             public $owner = false;
 
             // Page assets that can be registered and set by plugins (javascript, css, etc)
-            public $assets = array();
+            private static $assets = array();
 
             function init()
             {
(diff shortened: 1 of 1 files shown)


kylewm:
I'm going to push back on this a little because it seems very drastic. Wouldn't this mean any asset defined on any page would now be included at the top of *every* page?

mapkyca:
Shouldn't be... there's only ever one page per process.

mapkyca:
PHP being a script, not a separate program.. static variables don't persist beyond the lifetime of the script.

kylewm:
OK I see that. Could you explain how a plugin could set an asset on the wrong page?

mapkyca:
I'm not entirely sure tbh; it *looks* like there's a double instantiation of the page object going on (which is what I meant by "wrong page"), but I've not had a chance to track it down... this is a quick fix to get plugins working again while I dig deeper...

benwerd:
@mapkyca Any advance on this? I'm not sure I'm seeing this at my end.

mapkyca:
It's definitely a double creation problem.

Basically it's down to currentPage() in Idno.php combined with the plugin boot order; basically when plugin initialisation is triggered there is no current page. This error is silently squished however, since currentPage() will return a blank Page() object.

Unpacking this then we have a number of things going on:

1) a chicken and egg between plugins and pages - the current page needs to be created before plugins can perform page operations (related #235), but the page itself is created by a page handler which is often provided by a plugin and so needs to happen *after* plugins have been instantiated.

2) currentPage()'s fallback is to return a blank Page() object, which is destroyed as soon as the "real" page is created. All actions performed on this object are meaningless since it's state is not preserved, worse it's silently destroyed.



This last should *never* happen, so for example, instead of returning a bizarrely temporary object, Page() should be abstract and current page should return false, or perhaps better, throw a catchable "no page yet" exception.

Additionally, I think we're looking at needing an "after boot, but before render" hook on plugins where we can register things like assets (and input handlers as has previously cropped up)

mapkyca:
Closing in favour of #1356

PR #1305, opened 2016-02-26
add IndieSyndicate to plugin docs

kylewm:


benwerd:
I've just triggered a rebuild of the plugin docs to reflect this.

Issue #1247, opened 2016-02-16
Run db->optimize() manually


kylewm:
The call to `\Idno\Core\Idno::site()->db()->optimize();` when you first open the site configuration page takes a non-trivial amount of time to run (~15 seconds). Even after throttling it to once a day, it still irritates me, and if I didn't know why it was taking so long, I would be even more impatient.

I think I'd like to add a button to invoke it manually (also via the console script, so you could run it in a cron job) and stop doing it on page load. I could make this a configuration option if it's still advantageous to run it automatically for most sites.

benwerd:
Yeah. I think you're right - anything that lags page load should probably
be wiped.

I'm wondering if having a cron job hook wouldn't be a bad call (perhaps
with a token in the URL to prevent external attacks). You could then
automatically trigger some back-end tasks, like this, to run regularly. For
shared hosts that can't easily run cron jobs we could trigger via Convoy.

On Tue, Feb 16, 2016 at 11:26 AM, Kyle Mahan <notifications@github.com>
wrote:

> The call to \Idno\Core\Idno::site()->db()->optimize(); when you first
> open the site configuration page takes a non-trivial amount of time to run
> (~15 seconds). Even after throttling it to once a day, it still irritates
> me, and if I didn't know why it was taking so long, I would be even more
> impatient.
>
> I think I'd like to add a button to invoke it manually (also via the
> console script, so you could run it in a cron job) and stop doing it on
> page load. I could make this a configuration option if it's still
> advantageous to run it automatically for most sites.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/idno/Known/issues/1247>.
>



-- 
Ben Werdmuller
<http://goog_1933028737>
benwerd.com | werd.io

+1 (312) 488-9373

mapkyca:
+1 for the cron hook

mapkyca:
Another thing to consider, assuming the db is still available at that point (it should be), is to run the command on a shutdown hook. That way it'll run after the page has rendered.

kylewm:
Works for me!

mapkyca:
Another thing, the vendor service ping is also done on this thread, which can take a little time (and times out behind a firewall).

I wonder if we should do this a little smarter as well..

kylewm:
I'm still getting a long delay loading the /admin page the first time (tested w/ both nginx and apache). It looks like at some point, php's behavior changed to process shutdown functions before closing the connection to the user:

[...]

PR #1307, opened 2016-02-26
Shift Optimise to a service call.

mapkyca:

Introducing a service page hierarchy, moving optimise to a service call on admin/home

## Here's why I did it:

Because shutdown hooks no longer work.

Closes: #1247


benwerd:
:+1: :+1:

PR #1312, opened 2016-02-27
Fixing build

mapkyca:

Moved PHP 5.4 travis into allow_failing

## Here's why I did it:

2279f5f488d9d9823e807bd45c19c2a2447d0336 introduced a breaking change as CURLFile is only available in PHP 5.5 and above.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ env:
   
 matrix:
     allow_failures:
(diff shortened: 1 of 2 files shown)


mapkyca:
Closing in favour of #1315, which is a better solution.

Will keep #1313 open, as I think it still has merit.

PR #1314, opened 2016-02-27
Better exceptions

mapkyca:

Makes better use of the SPL Exception library; Follows best practice to not use Exception for everything, introduces some specific runtime exceptions to give better granularity.

## Here's why I did it:

It's not recommended to just throw \Exception() all the time, as it becomes much harder to gracefully handle specific problems. We should get in the habit of either using existing SPL functions, or extend the appropriate base to add our own.

### DIFF
--- a/Idno/Pages/Account/Settings/Following/Bookmarklet.php
+++ b/Idno/Pages/Account/Settings/Following/Bookmarklet.php
@@ -1,454 +1,190 @@
 <?php
 
     /**
-     * Utility methods for handling external web services
-     *
-     * @package idno
-     * @subpackage core
+     * Bookmarklet endpoint
      */
 
-    namespace Idno\Core {
+    namespace Idno\Pages\Account\Settings\Following {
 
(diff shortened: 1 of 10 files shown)


Idno/Pages/Account/Settings/Following/Bookmarklet.php
```diff
@@ -1,190 +1,454 @@
...
      */
 
-    namespace Idno\Pages\Account\Settings\Following {
+    namespace Idno\Core {
 
-        /**
-         * Default class to serve the following settings
-         */
-        class Bookmarklet extends \Idno\Common\Page
+        class Webservice extends \Idno\Common\Component
```
kylewm: What happened here? Looks like Bookmarklet.php was overwritten with Webservice.php?

mapkyca: Well, that would be my IDE going totally batshit.

What the actual.

Issue #1310, opened 2016-02-26
Overzealous translation of @-values to CURLFiles


kylewm:
## While trying to do this:

Call Webservice::post with a parameter `content="@benwerd blah blah blah"`

## I encountered this error:

> Error contacting micropub endpoint (0): couldn't open file "benwerd this is the best evidence I have so far that you *are* @PHP_CEO" 

![screenshot from 2016-02-26 12-07-44](https://cloud.githubusercontent.com/assets/950127/13364086/c2843e12-dc81-11e5-9b78-ee41874d49e0.png)

## Some other notes:
In a day of amusing errors, this is the most amusing so far.

kylewm:
suspect this was introduced in ef20dca3c18c3f81f7db8eb6bb38b460525a264b

mapkyca:
This is a problem with maintaining php 5.4 support.

kylewm:
Proposal: introduce a new class, maybe like `\Idno\Core\WebserviceFile`, that becomes the only sanctioned way to pass file handles to Webservice.

Then Webservice can convert that to @-syntax or CurlFile depending on the PHP version

mapkyca:
That would work, aye, and is probably the best solution in this instance, although I'm cautious of adding too many platform specific shims for obsolete platforms.

kylewm:
Oh dang, I blamed @mapkyca's CurlFile change, but this was *my* fault! db4b169009714bcb750160395a33aecd4764a658 removed the code that manually called http_build_query, which I now understand existed to work around the @-name issue.

I'm going to try setting `CURLOPT_SAFE_UPLOAD` (PHP 5.5+), which is supposed to prevent it from interpreting @'s.

kylewm:
I tested with #1315 and #1326 and confirmed that images and @-replies publish correctly now.

PR #1315, opened 2016-02-27
Adds WebserviceFile curl wrapper

mapkyca:

As per @kylewm's suggestion, introducing a webservice file wrapper around the various curl file formats

## Here's why I did it:

Removing the previous shim broke the build for > PHP 5.4, and while I think binning 5.4 is the longterm goal (#1313) this is a good solution to this specific issue.

Closes: #1310
Closes: #1312

### DIFF
--- a/Idno/Core/Webservice.php
+++ b/Idno/Core/Webservice.php
@@ -44,28 +44,20 @@
                 switch (strtolower($verb)) {
                     case 'post':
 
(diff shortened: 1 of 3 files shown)


kylewm:
This is excellent, thank you! :dancer:

mapkyca:
HTH :)

PR #1316, opened 2016-02-27
truncating long blog post 

physcocode:
truncated long blog post in homepage.
## Here's why I did it:
long blog post eat up a lot of space causing lot of scrolling on mobile devices which ultimately lead to poor user experience on mobile devices

### DIFF
--- a/IdnoPlugins/Text/Main.php
+++ b/IdnoPlugins/Text/Main.php
@@ -3,12 +3,27 @@
     namespace IdnoPlugins\Text {
 
         class Main extends \Idno\Common\Plugin {
+           
+           
+           function init()
+           {
+             parent::init();
+             if(empty(\Idno\Core\Idno::site()->config()->truncate_character)){
+             \Idno\Core\Idno::site()->config()->truncate_character=200; //Default Truncate length
+             }
+           }
+
 
             function registerPages() {
                 \Idno\Core\Idno::site()->addPageHandler('/entry/edit/?', '\IdnoPlugins\Text\Pages\Edit');
                 \Idno\Core\Idno::site()->addPageHandler('/entry/edit/([A-Za-z0-9]+)/?', '\IdnoPlugins\Text\Pages\Edit');
                 \Idno\Core\Idno::site()->addPageHandler('/entry/delete/([A-Za-z0-9]+)/?', '\IdnoPlugins\Text\Pages\Delete');
                 \Idno\Core\Idno::site()->addPageHandler('/entry/([A-Za-z0-9]+)/.*', '\Idno\Pages\Entity\View');
+                \Idno\Core\Idno::site()->addPageHandler('/admin/blog/?', '\IdnoPlugins\Text\Pages\Admin\Blog');
+                
+                \Idno\Core\Idno::site()->template()->extendTemplate('admin/menu/items', 'admin/menu/items/blog');
+
+
             }
         }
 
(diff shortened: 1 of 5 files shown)


physcocode:
checked on a PHP 5.5.28 installation works as expected yet failing travis tests in above PHP 5.4
with some SQL syntax error

kylewm:
I think this is the sort of thing where you can't make everyone happy with one stylistic choice. I personally wouldn't want articles truncated in my feed, but of course some people do.

Perhaps a per-site configuration option?

Also, there are better ways to truncate than simply chopping it at a certain number of characters. Maybe include only the first paragraph. Or at least, cut at a word boundary, so that you don't break in the middle of a word. Another option would be to have an explicit 'summary' field so someone could hand-author their own summary.

physcocode:
@kylewm it does not break a word it breaks only on whitespace. Well I agree this should be configurable , it doesn't effect much with desktop experience . But people who are using known says its an irritating experience on mobile device 

Maybe should we just truncate for mobile device ?

kylewm:
> it breaks only on whitespace.

Ah I see that now, sorry I misinterpreted.

physcocode:
for instance i do read @benwerd's werd.io its cool on my laptop . But when i read it on mobile device i would end up my life scrolling thorough long long blog that i've already read .
Something really has to be done about user experience on mobile devices thoughts @erinjo ?

physcocode:
@benwerd @kylewm can plugins have setting pages ?
here is a idea . text plugin could have setting page where user can decide 
1) Should blogs be truncated
2) Number of character to include while truncating.  
3) should break at whitespace our end of sentence i.e. "."
I am interested to submit a PR on this.

kylewm:
This is maybe part of a bigger UX problem that most indieweb sites (and readers) have IMHO... on one hand longer posts can take up too much vertical space; on the other, they tend to get lost in the stream, where maybe you spent a bunch of time researching writing an article, and a few photos of your breakfast and your cats later, it's gone.

werd.io handles this by only showing articles in the homepage, tantek.com by including links to "Recent Articles" in the sidebar.

physcocode:
@kylewm how can a plugin have a setting page in admin?

kylewm:
@physcocode you can use extendTemplate to add stuff to an existing template. If you want to add an entirely new page, then you generally extend the template that has the admin or account links, e.g. https://github.com/kylewm/KnownIndieSyndicate/blob/master/Main.php#L24

physcocode:
@kylewm truncating is now configurable :)

mapkyca:
In reply to a much earlier comment, travis builds are going to fail until #1315 is merged.

benwerd:
@erinjo has been asking for this since day one. I'm actually with @kylewm in that I wouldn't want it in the feed, and I kind of don't want it on my site. It's tempting to find a way to do this in pure JS: include a snip code in the blog post, that a piece of JS finds and truncates to. The readers, which don't parse JS, would see the whole post. And of course, you could simply turn the script off, in the way it's being toggled here.

physcocode:
@benwerd  I was trying to keep it simple , this feature was asked by many of my friends who are using known to host their personal site. So I thaught this can be a simple and pritty way 

I am also bit curious about having a summary , I was working on that too but now I will hold back as your preferring a JavaScript 

Should I close this PR if it is not relevant ? 
Or this need bit more iteration ?

physcocode:
I am releasing a separate plugin for this ! If it doesn't suit in the core !

PR #1317, opened 2016-02-27
move e-content to individual post templates

kylewm:

move e-content to individual post templates rather than wrapping the entire post in an e-content, which has the effect of e.g., including the post title inside the e-content

## Here's why I did it:

I don't like seeing the post title duplicated inside the post body in my feed reader. It's possible that this behavior is desireable for some reason, and I am open to suggestions.


kylewm:
hmm, having some CSS issues now.

benwerd:
Holding off on this; will test.

kylewm:
you're right to hold off :) expanding a little on the CSS issues: It's hard to get the font sizes right for both articles and statuses. The CSS is currently something like `.e-content p {font-size: 1.2em}`, which this change broke because statuses become `<p class="p-name e-content"> ... </p>`

benwerd:
Probably CSS shouldn't key off the microformats classes. I'm tempted to move them all entry-content or something else?

kylewm:
tbh, I'm pretty agnostic on whether to style based on mf classes, but it seems most on IRC are against it. `.article .e-content` and `.note .e-content` always seemed useful to me.

Also `entry-content` *is* a microformats (1) class. I think you'd have to add a new class (idno-content?) if you were serious about doing it.

The (hacky?) fix for my issue might just be to make the selector match the new hierarchy too `.e-content p, p.e-content`. 

removing ActivityStreamPost is causing me to move things around right now anyway, such that I'll need to redo this change after that one.

benwerd:
I'm an eejit. idno-content (heh, or known-content, two years after the name change) might be the way to do it.

kylewm:
Took me a while to come around, but I finally agree that styling based on e-content was the problem. I added a new div with class `idno-body`* that replaces e-content in the templates and CSS, and now everything should behave just as it did before.

* turns out idno-content was already taken, so I made a new class idno-body that represents the body of the post. content contains the post's header, body, and footer. I'm open to other naming suggestions as I find the hierarchy between -entry, -content, and -body pretty confusing.

kylewm:
The conflicts are resolved, man

benwerd:
:+1:

Issue #1309, opened 2016-02-26
Render syndication links on feed pages, or support ETags/If-Modified-Since


snarfed:
we've been working on optimizing bridgy to cut costs recently (snarfed/bridgy#578 , snarfed/bridgy#624), and one of its big sources of load currently is refetching home pages and permalinks to check for new/updated syndication links.

Known and WordPress are the two biggest CMSes bridgy sees by far. if you all were able to render synd links inside entries in feeds, that would let us skip fetching the permalinks, which would be a huge help.

alternatively, if you supported ETags/If-Modified-Since, we could use them to make them permalink fetches and processing much lighter. a bit more work for us, and not quite as big a win, but still something.

cc @kylewm. thanks in advance! hope you all are good. oh, and feel free to split this into two issues if you want, obviously.

benwerd:
Seems totally reasonable. Are you scanning the mf2 or the RSS feeds?

On Friday, February 26, 2016, Ryan Barrett <notifications@github.com> wrote:

> we've been working on optimizing bridgy to cut costs recently (
> snarfed/bridgy#578 <https://github.com/snarfed/bridgy/issues/578> ,
> snarfed/bridgy#624 <https://github.com/snarfed/bridgy/issues/624>), and
> one of its big sources of load currently is refetching home pages and
> permalinks to check for new/updated syndication links.
>
> Known and WordPress are the two biggest CMSes bridgy sees by far. if you
> all were able to render synd links inside entries in feeds, that would let
> us skip fetching the permalinks, which would be a huge help.
>
> alternatively, if you supported ETags/If-Modified-Since, we could use them
> to make them permalink fetches and processing much lighter. a bit more work
> for us, and not quite as big a win, but still something.
>
> cc @kylewm <https://github.com/kylewm>. thanks in advance! hope you all
> are good. oh, and feel free to split this into two issues if you want,
> obviously.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/idno/Known/issues/1309>.
>


-- 
Ben Werdmuller
<http://goog_1933028737>
benwerd.com | werd.io

+1 (312) 488-9373

snarfed:
thanks! mf2, from the rendered pages, usually starting from the home page.

benwerd:
OK. So would mf2 syndication links in a sub div for each entry hidden with
display:none be acceptable?

On Friday, February 26, 2016, Ryan Barrett <notifications@github.com> wrote:

[...]

PR #1318, opened 2016-02-27
remove rel=syndication when rendering syndication links in the h-feed

kylewm:

remove rel=syndication when rendering syndication links in the h-feed

## Here's why I did it:

rel=syndication is page-scoped so they should only be used when we're on
the post's permalink


kylewm:
fixes #1309 (again)

PR #1323, opened 2016-02-28
Use youtube privacy enhanced player by default

mapkyca:

Changed the default embed url to youtube-nocookie.com

## Here's why I did it:

Privacy is good, plus the current embed would explicitly strip privacy from someone who pastes a no-cookie link.


benwerd:
I had no idea this existed! Cool.

PR #1334, opened 2016-03-01
Strip % characters in slug, closes #1331

### LINKED ISSUE #1331: Slugs allow percentages (and therefore break)
## While trying to do this:

Saving a post that had a % in the title

## I encountered this error:

The % made it to the slug, which of course triggers the browser to think of it as an encoded value, causing the post to be lost and breaking everything.

## Some other notes:

This was reported by a Known Pro user and duplicated by me.

mapkyca:

Stripped % characters in stubs, %encoding should already have been urlencoded.

## Here's why I did it:

Reported problems with % chars in stub


benwerd:
Literally just sat down to do this. Thank you for beating me to it!

mapkyca:
HTH :)

PR #1346, opened 2016-03-06
Completed licensing information for included libs

lcorbasson:
* Added precisions on the BSD licenses.
* Added missing licensing info.

## Here's why I did it:
It makes things clearer for reuse of Known or its components.


benwerd:
Vital. Thank you.

PR #1347, opened 2016-03-07
Unicode slug fix

### LINKED ISSUE #1343: Non-Ascii character in title force "untitled" in permalink
If I use a non-ascii character like for example german umlauts (äöü) in a title the permalink is build with "untitled".

I propose to replace non-ascii characters by a hyphen.

kylewm:

- new unit test for prepareSlug with some simple cases and some umlauts (both inside and outside of the slug length)
- allow % characters when creating the slug; prevents it from converting %C3%BC to C3BC
- specify 'UTF-8' encoding in mb_strtolower

## Here's why I did it:

On some versions of PHP, `mb_strtolower` without a specific converts ü to \343\247, which is some other encoding that breaks everything and erases the slug.

Fixes #1343

### DIFF
--- a/Idno/Common/Entity.php
+++ b/Idno/Common/Entity.php
@@ -627,7 +627,7 @@
                 $slug = preg_replace_callback("/([\p{L}]+)/u", function ($matches) {
                     return rawurlencode(($matches[1]));
                 }, $slug);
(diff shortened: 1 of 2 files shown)


benwerd:
Aha! Explained and fixed before I had even poured my coffee. Thank you yet again @kylewm!

Issue #1352, opened 2016-03-08
IndiePub syndicate-to-expanded property is blank on hosted Known


kylewm:
## While trying to do this:

Fetch `?q=syndicate-to` from the micropub endpoint
## I encountered this error:

syndicate-to looks good, but syndicate-to-expanded is empty, which means Woodwind draws ugly syndication buttons for @kevinmarks:

```
{
    "syndicate-to": [
        "twitter::kevinmarks",
        "facebook::10154585682935425",
        "facebook::863215510399318",
        "facebook::688447421193285",
        "facebook::159878817395294",
        "facebook::36123201464"
    ],
    "syndicate-to-expanded": []
}
```
## Some other notes:

This only seems to affect hosted Known. Tested known.kevinmarks.com and kylewm.withknown.com. Briefly reading the code, I don't even really see how this result is possible as both `Syndication::getServiceAccountStrings` and `Syndication::getServiceAccountData` iterate over the same structure.

benwerd:
I'm guessing this might also affect Convoy users? It's because the account data isn't actually stored on the site, but is registered in the Convoy database / server.

kylewm:
good lord, you'd think I would have remembered that since I wrote that code in IndiePub. thanks Ben.

benwerd:
On my list for today is updating this mechanism (partially so I can get the latest code on withknown.com). Is there something I should be looking at here in the same swoop?

kylewm:
It'd be awesome if there were a way to request the syndication endpoints as JSON from `hub/user/syndication`. I don't think there's quite enough info in the HTML to reconstruct the structured data version of the endpoints.

https://github.com/idno/Known/blob/master/IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php#L19

kevinmarks:
did you and @aaronpk converge this? the wiki has multiple suggestions on
how to do this
http://indiewebcamp.com/Micropub-brainstorming#Discovering_Supported_Syndication_Targets

kylewm:
@kevinmarks nope, I'm just using what Known gives me which is the "syndicate-to-expanded" bit from that document.

benwerd:
Reproduced and confirmed that the data isn't trickling down from the hub (although locally-saved account data, like webhook data, is).

kevinmarks:
so, what can I do to mitigate this? anything? my woodwind buttons are still ugly

kylewm:
[...]

PR #1357, opened 2016-03-10
IndiePub needs to call static function as static

mapkyca:

Call to findUserForToken() should be static, otherwise it's undefined.

## Here's why I did it:

It should be static.

Possibly related to #1352


kylewm:
:+1: Whoa, how did that ever work! This will probably fix an issue @rmendes was having with Micropub from Woodwind too (that we have since worked around, but will be nice to have a proper fix)

PR #1367, opened 2016-03-15
ignore webmentions where the source is a feed

### LINKED ISSUE #1358: inexistent Reply notifications to my own post ? what's the point??
## While trying to do this:
I post something new at my site https://www.rmendes.net/2016/why-do-i-get-reply-notifications-from-withknown-on-pages
## I encountered this error:
then i get several notifications that I replied to my own post, when I didn't & nothing show replies under the aforementioned post in the notification area.
screenshot_70

## Some other notes:
Notice that notifications reply are about a static page on the same site that I would never reply to
I'm not even sure where this is coming from.  European Politics and Belgian Politics are 2 static pages running a rebelmouse aggregation on these 2 topics.

kylewm:

- add unit tests for `Entity::addWebmentions`
- examine incoming webmention source to see if it looks like a feed (has > 1 h-entry)
- discard webmentions from pages that look like feeds
- refactored Entity::addWebmentionItem to remove duplicate code for parsing h-cards

## Here's why I did it:

If you send a webmention to yourself wtih `source=http://yoursite.com/` and `target=http://yoursite.com/2016/a-permalink`, it currently sends yourself a reply from every post on the feed to that target, which looks crazy (see #1358 for screenshots)

fixes #1358

### DIFF
--- a/Idno/Common/Entity.php
+++ b/Idno/Common/Entity.php
@@ -1612,7 +1612,7 @@
              *
              * @param string $source The source URL
              * @param string $target The target URL (i.e., the page on this site that was pinged)
-             * @param string $source_content The source page's HTML
+             * @param array $source_content The Webservice response from fetching the source page
              * @param array $source_mf2 Parsed Microformats 2 content from $source
              * @return bool Success
              */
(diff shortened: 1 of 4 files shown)


rikmendes:
i'm just waw :)

Issue #691, opened 2015-01-08
Known needs a queue


benwerd:
A queue system will improve processing of:
- syndication
- outgoing webmentions
- incoming webmentions

This will make the platform faster and more resilient overall.

However, most self-hosted users won't technically be able to use queues. Therefore, I'm tempted to create an interface that _allows_ you to use your own local queue server, but also can use our infrastructure, in case you're on a host that doesn't support this.
##

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/idno/issues/691" class="u-in-reply-to"></a>Aye... just define the interface, have the an implementation which implements it synchronously, but have dispatchers for things like amazon's messaging server thingy etc....

Phyks:
Hmm, what would be the behaviour of such a queue system ? As I understood it so far, it could be a local queue, in local database, of outgoing queries that could not be processed, to process them once again in the future.

But your remark about self-hosting seems to imply that it should be running on another server, a bit on a CDN model?

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/691" class="u-in-reply-to"></a>So, for syndication and outgoing webmentions, I think it might be useful (especially if withknown.com provided a hosted service, as has previously been discussed), to bake in some techniques to mitigate mass interception (as, I have mentioned before, is IETF best practice now)

We have previously discussed as part of #203, that it might be advantageous to use some techniques to obfuscate a user's social graph, since this is often more valuable to an attacker than message content and is harder to protect (should be noted that it is still possible to extract, but it's just not as easy). We could, for example, route mentions over tor etc.

Trouble is, if you're watching that network, you can perform statistical analysis based on the traffic being sent - size of packet, time it arrived, etc. This is called a confirmation attack.

[...]

PR #1391, opened 2016-03-29
add a Queue interface and default (sync) implementation

kylewm:

Add a Queue interface and default (sync) implementation. Move some events (webmentions, PuSH ping, syndication) to the queue first.

## Here's why I did it:

- allow some events to send to a queue instead of triggering immediately
- EventQueue is super generic to hopefully allow experimentation, avoid
  specifying exactly how a queue dispatcher should work

This PR continues my thought process from #691, I don't necessarily need it merged so much as I wanted to continue the conversation... FWIW, I have a proof of concept plugin that uses Gearman running locally, but it might be a little while before I can release it.

### DIFF
--- a/Idno/Common/Entity.php
+++ b/Idno/Common/Entity.php
@@ -1677,10 +1677,9 @@
                         $this->save();
 
                         if ($return && $this->isReply()) {
-                            $webmentions = new Webmention();
                             if ($reply_urls = $this->getReplyToURLs()) {
                                 foreach ($reply_urls as $reply_url) {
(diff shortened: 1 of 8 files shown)


rikmendes:
could this be in some way a future Buffer like queuing system, i'm thinking Syndicate Elsewhere later, even if I do publish the post.

kylewm:
woo! I should mention I have a draft async queue implementation here:

https://github.com/kylewm/KnownSimpleQueue

It stores jobs in the regular datastore, and executes them via the `worker.php` worker running in an external process.

PR #1395, opened 2016-04-07
bugfix: check that ContentType is not false before inspecting it

### LINKED ISSUE #1394: Call to a member function getEntityClass() on boolean
## While trying to do this:
Editing an existing page for a typo/modification
## I encountered this error: (when saving the page)
Fatal error: Call to a member function getEntityClass() on boolean in /rmendes.net/Idno/Core/PubSubHubbub.php on line 253
Oh no! Known experienced a problem!

Known experienced a problem with this page and couldn't continue. The technical details are as follows:

Fatal Error: /rmendes.net/Idno/Core/PubSubHubbub.php:253 - "Call to a member function getEntityClass() on boolean", on page www.rmendes.net/pages/my-cv
## Some other notes:
I can confirm this happens when saving the page. 
the page get saved even though the crash bug happens without returning 
to the site when saving finished. 
screenshot_116

kylewm:

Check that $entity->getContentType() returns something before trying to access it.

## Here's why I did it:

$entity->getContentType() returns false or StaticPage entities.

Fixes #1394

### DIFF
--- a/Idno/Core/PubSubHubbub.php
+++ b/Idno/Core/PubSubHubbub.php
@@ -250,7 +250,7 @@
                     $homepage_types   = \Idno\Core\Idno::site()->config()->getHomepageContentTypes();
                     $type_in_homepage = false;
                     if ($object instanceof Entity) {
-                        if (in_array($object->getContentType()->getEntityClass(), $homepage_types)) {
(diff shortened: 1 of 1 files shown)


benwerd:
👍  it'll probably return false on more in the future.

Issue #1396, opened 2016-04-07
A reply to a post that is not POSSEd to twitter should include the replied-to post's URL when POSSEd to twitter


kevinmarks:
## While trying to do this:

When I use Known to reply to a post in Woodwind or using the Known bookmarklet, I don't know if the linked-to post will be sent to twitter.
## I encountered this error:

If the post I'm replying to was POSSEd to twitter, then Known sets the in-reply-to correctly on the POSSEd copy. If it wasn't, then there is no cue that ti was a reply to anything.
## Some other notes:

This post was not POSSEd to twitter:
http://known.kevinmarks.com/2016/time-is-hard---this-post-was-from-the-future
This reply to it was:
http://known.kevinmarks.com/2016/kylewmhan-if-i-reply-to-this-on-known-and-posse
https://twitter.com/kevinmarks/status/718181774816219136

Note that the twitter version has no indication it was a reply to anything

This post, replying to the 2nd one (which was POSSEd to twitter):
http://known.kevinmarks.com/2016/kylewmahan-replying-to-a-possed-post-does-set-the-in-reply-to

does set the in-reply-to on twitter, showing the thread:
https://twitter.com/kevinmarks/status/718182609000341505

So, when there is a POSSE to twitter of an in-reply-to post, and the post being replied to does not have a twitter POSSE copy, Known should put the URL of the post being replied to in the twitter POSSE copy.

kylewm:
fwiw, we handled this in Red Wind and Bridgy but simply *not letting* you POSSE a reply to Twitter if we can't find the ID of the tweet being replied to.

in silo.pub, I do something more like your suggestion -- construct a tweet of the form:

```
"Re: https://url-being-reply-to, I disagree with you!"
```

Although tbh, I never want to tweet something that looks like that :P

kevinmarks:
Right, but when I'm reading an indie post in woodwind, I don't know if it
has a POSSE copy, so I am not sure what will happen when I reply there; the
'reply without context' is odd as well.

snarfed:
lots more discussion on the use case and tradeoffs in snarfed/bridgy#362 (and another request in snarfed/bridgy#622).

benwerd:
I'm not in favor of invisibly syndicating text to Twitter that you didn't explicitly write.

[...]

PR #1400, opened 2016-04-09
improve UI when replying to an original that is syndicated to Twitter

kylewm:

- update php-mf2 to latest version (old version did annoying, albeit reasonable,
  thing returning stdClass for empty "rels" array)
- improve Webmention::addSyndicatedReplyTargets to search u-syndication
  in addition to rel-syndication
- pre-fill the twitter @-name of the person we are replying to (e.g. if I reply to
  a post on known.kevinmarks.com that is syndicated to twitter, it will
  prefill with "@kevinmarks ")
- pass syndication URLs to the preselect hook so the Twitter plugin
  can preselect syndication

## Here's why I did it:

see discussion in https://github.com/idno/Known/issues/1396#issuecomment-207215698

### DIFF
--- a/Idno/Core/Webmention.php
+++ b/Idno/Core/Webmention.php
@@ -94,21 +94,35 @@
              * adds and rel="syndication" URLs in the target to the array
              * @param $url
              * @param array $inreplyto
+             * @param array $response (optional) response from fetching $url
              * @return array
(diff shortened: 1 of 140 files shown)


benwerd:
OMG

PR #1402, opened 2016-04-12
upgrade MentionClient to pass webmention.rocks tests

### LINKED ISSUE #1401: Webmentions to webmention.rocks don't all show up
## While trying to do this:
Link to all the webmention.rocks tests from [this post](http://known.kevinmarks.com/2016/test-page-for-pinging-webmentionrocks-from) and send webmentions
## I encountered this error:
Only 2 showed up. 
Specifically 4 and 6 - which implies Link headers and relative URLs need checking

Test 1 - HTTP Link header, unquoted rel, relative URL
Test 2 - HTTP Link header, unquoted rel, absolute URL
Test 3 - HTML <link> tag, relative URL
Test 4 - HTML <link> tag, absolute URL
Test 5 - HTML <a> tag, relative URL
Test 6 - HTML <a> tag, absolute URL
Test 7 - HTTP Link header with strange casing
Test 8 - HTTP Link header, quoted rel
## Some other notes:
I was surprised that none of the Link headers worked

kylewm:

- MentionClient needed a little tweak to support rel=webmention style Link headers (no quotes)
- for better or worse, MentionClient has changed its interface a bunch since we last upgraded it, so I had to rename some methods and such

## Here's why I did it:

fixes #1401  (evidenced https://kylewm.com/2016/04/trying-webmentionrocks-from-known-2)


kylewm:
(I would recommend holding off until after the next release. Lots of churn, not a lot of testing yet)

PR #1406, opened 2016-04-17
don't double encode unicode characters in setSlugResilient

kylewm:

- remove redundant call to prepareSlug at the beginning of setSlugResilient

## Here's why I did it:

- calling prepareSlug twice meant that unicode characters got
  converted from ü to %C3%BC to %25c3%25bc which was bad!

### DIFF
--- a/Idno/Common/Entity.php
+++ b/Idno/Common/Entity.php
@@ -594,9 +594,17 @@
                 $slug = preg_replace("/([^A-Za-z0-9%\p{L}\-\_ ])/u", '', $slug);
                 $slug = preg_replace("/[ ]+/u", ' ', $slug);
                 $slug = implode('-', array_slice(explode(' ', $slug), 0, $max_pieces));
-                $slug = rawurlencode($slug);
 
-                $slug = substr($slug, 0, $max_chars);
+                // trim the source string until the encoded string is <= max_chars
+                for ($chars = $max_chars ; $chars >= 0 ; $chars--) {
+                    $truncated = mb_substr($slug, 0, $chars, 'UTF-8');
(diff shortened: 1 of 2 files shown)


kylewm:
Please consider merging this *before* the release.

rikmendes:
is this related to some of my URL's to break when shared on FB for exemple ? (not when POSSed)
but manual sharing this : https://www.rmendes.net/2016/le-proc%25c3%25a8s-okc-spatz-cest-une-tentative-par-la-d%25c3%25a9fense return a "Gone baby, gone" on facebook. even though the link seems to work...kinda strange..

kylewm:
I tacked on another change -- when truncating encoded slugs to a certain length (e.g., 255 chars), it was truncating in the middle of a byte sometimes so you'd have like `...%D0%B4%D0%B0-%D` where `%D` is just part of a bytes

So now it truncates before encoding, iteratively trying shorter and shorter source strings until the encoded string fits in the max characters bounds.

kylewm:
@rikmendes I don't think it's related :/

kylewm:
@rikmendes I was wrong, you were right, it does look related! Facebook "fixes" the double encoding `%25c3%25a8` changing it back to `%C3%A8` which Known doesn't know about.

So this change should fix that issue for you going forward, though old posts will still have the double encoded slug.

rikmendes:
@kylewm interesting..because when the that link got POSSED it was a status post type http://rmendes.net/s/3VwNM https://www.facebook.com/1619845158237630/posts/1723062681249210  all fine to facebook - so is there 2 ways the url is generated ? or I mean processed ?

PR #1426, opened 2016-05-05
Add pronouns to Known profiles

benwerd:

I added pronoun support for Known. Right now it needs to be manually added by setting `enable_pronouns=true` in `config.ini`. Eventually, once we've arrived at a good way to display this, we should remove the config check and just make it live.
## Here's why I did it:

Pronouns are not a given and part of owning your own identity is having the right to state yours clearly.

### DIFF
--- a/Idno/Entities/User.php
+++ b/Idno/Entities/User.php
@@ -188,6 +188,48 @@
             }
 
             /**
+             * Retrieve this user's oblique pronoun
+             * @return string
+             */
+            function getPronounOblique()
+            {
+                return !empty($this->profile['pronoun-oblique']) ? $this->profile['pronoun-oblique'] : '';
+            }
+
+            /**
+             * Retrieve this user's nominative pronoun
+             * @return string
+             */
+            function getPronounNominative()
+            {
+                return !empty($this->profile['pronoun-nominative']) ? $this->profile['pronoun-nominative'] : '';
+            }
+
+            /**
+             * Retrieve this user's possessive pronoun
+             * @return string
+             */
+            function getPronounPossessive()
+            {
+                return !empty($this->profile['pronoun-possessive']) ? $this->profile['pronoun-possessive'] : '';
+            }
+
+            /**
+             * Retrieve this user's pronouns
+             * @return array
+             */
+            function getPronoun()
+            {
+                $pronoun = [
+                    'oblique'    => $this->getPronounOblique(),
+                    'nominative' => $this->getPronounNominative(),
+                    'possessive' => $this->getPronounPossessive()
+                ];
+                if (!sizeof(array_filter($pronoun))) return [];
+                return $pronoun;
+            }
+
+            /**
              * Retrieves user by email address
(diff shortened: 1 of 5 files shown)


templates/default/entity/User/profile/pronouns.tpl.php
```diff
@@ -0,0 +1,11 @@
+<p class="p-pronoun pronoun small">
```
kylewm: sorry to be a microformats prescriptivist, but I don't think p-pronoun is a recognized/documented property for [h-card](http://microformats.org/wiki/h-card#Properties).

kylewm:
Tried this out on https://known.kylewm.com/profile/kyle, lgtm! 

And the parsed version looks nice too: http://mf2.kylewm.com/?url=https%3A%2F%2Fknown.kylewm.com%2Fprofile%2Fkyle&parser=html5lib (with the exception of `p-pronoun` mentioned above)

Idno/Entities/User.php
```diff
@@ -188,6 +188,48 @@ function getHandle()
...
+
+            /**
+             * Retrieve this user's pronouns
+             * @return array
+             */
+            function getPronoun()
+            {
+                $pronoun = [
+                    'oblique'    => $this->getPronounOblique(),
+                    'nominative' => $this->getPronounNominative(),
```
kylewm: I'm used to seeing nominative first ("he/him/his" rather than "him/he/his")

mapkyca:
Yep.

Question, how will this connect with i18n?

kylewm:
At HWC we discussed maybe using properties like `p-x-pronoun-en-nominative` (or `p-x-en-pronoun-nominative`?)

Another possibility is determining the language of the containing element or document (using the lang attribute), but then {nominative, oblique, possessive} are almost certainly not sufficient for many other languages

I know @voxpelli, @kevinmarks, and others have discussed how to mark up pages that have multiple languages within a single same page, but I'm not up to speed on those discussions.

voxpelli:
This is the related parsing brainstorm around languages: http://microformats.org/wiki/microformats2-parsing-brainstorming#Parse_language_information

As there is a proper `lang="en"` attribute in HTML that already conveys the needed information I feel that something like `<span class="p-x-pronoun-nominative" lang="en">` be a much nicer solution than to introduce some kind of language into the class itself.

This could perhaps be a good example to document at the above page so that we can push the brainstorm around how to solve that issue further towards a solution?

kylewm:
Thanks @voxpelli, I've added the brainstorming section you suggested: http://microformats.org/wiki/microformats2-parsing-brainstorming#Pronouns_in_different_languages

I don't think we should block inclusion in Known on resolving those questions... The real-world publishing experience we get here will help inform the spec.

mapkyca:
This has been open for a while now... votes to merge?

This'll have a visual change, so not just going ahead and doing it...

mapkyca:
This doesn't seem to be going anywhere, so closing.

PR #1428, opened 2016-05-05
Reset object's attachment array after deleting attachment files.

mapkyca:

Reset the ->attachments array

## Here's why I did it:

Deleting object attachments removed the files but not the array, resulting in duplicates / broken images when displaying.


mapkyca:
Bump.

You'll probably need this.

benwerd:
👍 thank you!

PR #1430, opened 2016-05-09
Per object class edit menu

mapkyca:

Per object class edit menu

## Here's why I did it:

I needed a simple way to add edit options on specific classes.


templates/default/content/edit.tpl.php
```diff
@@ -8,6 +8,7 @@
         <a href="<?=$vars['object']->getEditURL()?>" class="edit">Edit</a>
         <?=  \Idno\Core\Idno::site()->actions()->createLink($vars['object']->getDeleteURL(), 'Delete', array(), array('method' => 'POST', 'class' => 'edit', 'confirm' => true, 'confirm-text' => 'Are you sure you want to permanently delete this entry?'));?>
 
+	<?= $this->draw('content/entity/' .  substr(get_class($vars['object']), strrpos(get_class($vars['object']), '\\') + 1). '/edit'); ?>
```
benwerd: Can you talk me through this?

mapkyca: It provides an easy way to add extra controls to the object edit /delete bar but on a per-object basis.

If for example you create a custom object type, but you want a custom context specific menu option on that bar. This lets you do it in an easy(er) way.

PR #1431, opened 2016-05-10
Notice as exception

mapkyca:

* Known logger now intercepts PHP error messages
* Introducing *pedantic_mode* which will convert all PHP error messages to ErrorException()

## Here's why I did it:

* Notice errors are verbotten, using pedantic_mode during development will make it easier to enforce this discipline.

### DIFF
--- a/Idno/Core/Config.php
+++ b/Idno/Core/Config.php
@@ -58,6 +58,7 @@
                 'bypass_fulltext_search' => false,
                 'permalink_structure'    => '/:year/:slug',
                 'single_user'            => true,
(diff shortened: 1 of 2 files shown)


benwerd:
Why do we need this? I'm tempted to say we should just display notices? But I'm likely missing something :)

mapkyca:
For one: Displaying notices can be easily squished, exceptions are a hard failures which can't be ignored.

This is similar to a common compiler flag "warnings as errors". Obvs, it's not on by default.

PR #1433, opened 2016-05-13
Removing "denied!"

benwerd:

Removed "denied!" from denied pages.

## Here's why I did it:

It's an off-putting user experience. With an exclamation mark and everything!

### DIFF
--- a/Idno/Common/Page.php
+++ b/Idno/Common/Page.php
@@ -614,14 +614,7 @@
             function createGatekeeper()
             {
                 if (!\Idno\Core\Idno::site()->canWrite()) {
-                    $this->deniedContent();
-//                    $this->setResponse(403);
-//
-//                    if (!\Idno\Core\Idno::site()->session()->isAPIRequest()) {
-//                        $this->forward();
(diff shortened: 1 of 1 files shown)


Idno/Common/Page.php
```diff
@@ -614,27 +614,20 @@ function flushBrowser()
...
                 if (!\Idno\Core\Idno::site()->canWrite()) {
-                    $this->deniedContent();
-//                    $this->setResponse(403);
-//
-//                    if (!\Idno\Core\Idno::site()->session()->isAPIRequest()) {
-//                        $this->forward();
-//                    } else {
-//                        $this->deniedContent();
-//                    }
+                    $this->deniedContent('');
```
mapkyca: Can't we leave setting title to "" to the default?

mapkyca:
If we're going to pass a title as parameter, I'd suggest doing the same with the other xxxxContent() methods, for consistency.

benwerd:
Now fixed, with @mapkyca's suggestion merged in.

PR #1438, opened 2016-05-14
Fixing export on Mongo

### LINKED ISSUE #1437: Export fails
## While trying to do this:

Attempting to generate an export

## I encountered this error:

The export process runs but no archive is generated. May be related to #1436.

## Some other notes:

This is on Mongo, so maybe it's attempting to store using gridFS..?

mapkyca:

Some extra debug to the export; Isolated and fixed infinite loop when exporting a mongo based database.

## Here's why I did it:

Attempt to track down why export is failing (#1437)

Closes #1437

### DIFF
--- a/Idno/Common/Page.php
+++ b/Idno/Common/Page.php
@@ -117,6 +117,9 @@
             {
                 $this->setResponse(500);
                 http_response_code($this->response);
+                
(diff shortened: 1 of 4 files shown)


mapkyca:
Updated definition, this patch does more than debug :)

/cc @benwerd

PR #1441, opened 2016-05-17
PHP 5.4 compatibility

### LINKED ISSUE #1440: Cannot get static pages to work while upgrading to latest version
## While trying to do this: 

upgrade to the latest version of Known (0.9.2)

## I encountered this error:

The whole site doesn't work and error is

PHP Fatal error:  Can't use function return value in write context in /public_html/known/IdnoPlugins/StaticPages/Main.php on line 51

## Some other notes:

If i a remove the above folder (i.e. StaticPages) then the site works, but of course without static pages functionality.

would you know how to resolve this?

mapkyca:

Passing function result to variable before passing to empty().

## Here's why I did it:

Using function result in empty() isn't supported in PHP < 5.5, and while php 5.4 is no longer officially supported, this was a simple fix.

Closes #1440

### DIFF
--- a/IdnoPlugins/StaticPages/Main.php
+++ b/IdnoPlugins/StaticPages/Main.php
@@ -48,7 +48,8 @@
 
                 if (\Idno\Core\Idno::site()->session()->isLoggedIn()) {
                     if (\Idno\Core\Idno::site()->session()->currentUser()->isAdmin()) {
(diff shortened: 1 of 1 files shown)


kylewm:
oops, sorry for reopening that issue, didn't realize you already fixed! this lgtm

Issue #1457, opened 2016-05-21
Convoy infinity loop loading


hugoroy:
## While trying to do this:
1. In Admin > Connect services
2. Subscribe to Convoy
## I encountered this error:

I got redirected to my domain admin panel on "Connect Services" --- then it only loads infinitely something from domains.withknown.com

Tried the link by email http://MyDomain/convoy/token/?convoy_token=XXXXX (same result)
## Some other notes:

Glad to test Known again :-)

kylewm:
At one point did this redirect loop thing happen -- Were you able to successfully sign up for Convoy (like give your credit card number and everything) but then when it redirected to your site, it wouldn't load the iframe correctly? Or did the loop happen before it got that far?

hugoroy:
I got redirected back to the site after giving the payment info and on the admin panel, on my domain, I get infinite loading

kylewm:
Dang, well, the 302 redirect from "connect" seems strange. That doesn't happen in mine:

![screenshot from 2016-05-21 13-47-10](https://cloud.githubusercontent.com/assets/950127/15450783/aec5be48-1f5a-11e6-9c0f-018b4b58caed.png)

This looks like a job for @benwerd!

mapkyca:
<a href="https://github.com/idno/Known/issues/1457" class="u-in-reply-to"></a>Anything in the error log?

Looking at the code there's only a couple of places that'd bounce to referrer, and that's if there's no hub defined or it couldn't connect to the remote hub.

The former is worth checking that you've not somehow overridden the hub in config, and that it's reachable.

Anything more than that is up to @benwerd, but it might be worth improving the debug in the client side convoy plugin..

hugoroy:
From my side, I couldn't see anything in the console.

mapkyca:
<a href="https://github.com/idno/Known/issues/1457" class="u-in-reply-to"></a>Do you have a hub defined?

mapkyca:
<a href="https://github.com/idno/Known/issues/1457" class="u-in-reply-to"></a>... Not pubsub, but known hub.... which you should have defined, unless there's a bug in config.

What does it say in diagnostics?

hugoroy:
I know about the PubSubHubbub hub which is https://withknown.superfeedr.com/ (the default). What other hub are you talking about?

This is a fresh install from the zip available on the website.

kylewm:
[...]

PR #1458, opened 2016-05-22
Refs #1457: Adding some debug to Convoy connect

mapkyca:

Added some debug to convoy connect

## Here's why I did it:

Try and see what is actually going on with #1457


kylewm:
lgtm

PR #1459, opened 2016-05-22
Handle geo location failures gracefully

mapkyca:

* Debug to geo location checkin code
* Error handler now displays error messages
* When location fails, we initialise the map to a default location and allow manual entry

## Here's why I did it:

Location failures would give an infinite loop

### DIFF
--- a/IdnoPlugins/Checkin/checkin.js
+++ b/IdnoPlugins/Checkin/checkin.js
@@ -1,68 +1,70 @@
 var CheckinMap, CheckinMarker;
 
-    function initMap(latitude, longitude) {
-        $('#geoplaceholder').hide();
-        $('#geofields').slideDown();
+function initMap(latitude, longitude) {
+    $('#geoplaceholder').hide();
+    $('#geofields').slideDown();
 
-        var map = L.map('checkinMap').setView([latitude, longitude], 15);
-        var layer = new L.StamenTileLayer("toner-lite");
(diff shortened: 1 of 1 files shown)


IdnoPlugins/Checkin/checkin.js
```diff
@@ -1,67 +1,70 @@
...
+	    longitude = position.coords.longitude;
+    queryLocation(latitude, longitude, initMap.bind(this, latitude, longitude));
+}
 
-    function handleErrorFromNavigator() {
-    }
+function handleErrorFromNavigator(err) {
+    addErrorMessage(err.message);
+    //$('#geoplaceholder').html('<p>Could not find your location: '+err.message +'</p>');
+    initMap("51.478791", "-0.01068"); // Allow failures to allow manual entry. Default to the center of the universe.
```
kylewm: ha! :+1:

IdnoPlugins/Checkin/checkin.js
```diff
@@ -1,67 +1,70 @@
 var CheckinMap, CheckinMarker;
-$(function () {
```
kylewm: why move this down? did these functions need to be in the global scope?

mapkyca: Probably not, I'll change that back.

kylewm:
Lgtm!

PR #1466, opened 2016-05-28
Closes #1465: Allows the commonly used .@ reply notation

### LINKED ISSUE #1465: Twitter activation doesn't handle .@ notation
## While trying to do this:

Post the following tweet ".@ScottishPEN @timberners_lee, @TheEconomist Already is - ECHR Art. 8-11, encryption protects those in the digital age."

## I encountered this error:

@timberners_lee & @TheEconomist are activated, but @scottishPen are not, due to the commonly used "." notaion

mapkyca:

Modified regex to allow .@ username notation

## Here's why I did it:

It's common practice to use .@user at the start of a tweet to make a public reply, however the local copy didn't activate the username.

### DIFF
--- a/Idno/Core/Template.php
+++ b/Idno/Core/Template.php
@@ -418,7 +418,7 @@
                         if (is_array($in_reply_to))
                             $in_reply_to = $in_reply_to[0];
 
-                        $r = preg_replace_callback('/(?<=^|[\>\s\n])(\@[\w0-9\_]+)/i', function ($matches) use ($in_reply_to) {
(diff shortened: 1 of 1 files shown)


kylewm:
LGTM :+1:

chrisaldrich:
&lt;sarkmark&gt;Glad to see this fixed just before Twitter kills the .@ reply construct: https://blog.twitter.com/express-even-more-in-140-characters &lt;/sarkmark&gt;

PR #1472, opened 2016-06-06
Create htaccess-2.4.dist

rickard2:

Added htaccess dist file for Apache 2.4

## Here's why I did it:

The default .htaccess file does not work with Apache 2.4


benwerd:
Thank you!

Issue #1477, opened 2016-06-22
Uncaught TypeError: expected Exception, Error given (in Idno/Common/Page.php)


rosetree:
## While trying to do this:
Adding a new bookmark or saving an edited post.
## I encountered this error:
````
Fatal Error: /var/www/known/Idno/Common/Page.php:116 - "Uncaught TypeError: Argument 1 passed to Idno\Common\Page::exception() must be an instance of Exception, instance of Error given, called in /var/www/known/Idno/Common/Page.php on line 71 and defined in /var/www/known/Idno/Common/Page.php:116
Stack trace:
#0 /var/www/known/Idno/Common/Page.php(71): Idno\Common\Page->exception(Object(Error))
#1 [internal function]: Idno\Common\Page->Idno\Common\{closure}(Object(Error))
#2 {main}
  thrown", on page localhost/known/2016/slug-of-my-private-post
````
## Some other notes:
This happened first after I upgraded from PHP 5.6 to PHP 7.0 on Ubuntu 16.04.

PR #1478, opened 2016-06-23
Pages errors

mapkyca:

Removed \Exception type constraint.

Removed a notice error.

## Here's why I did it:

PHP7 has different type hierarchy.

### DIFF
--- a/IdnoPlugins/StaticPages/templates/default/staticpages/admin.tpl.php
+++ b/IdnoPlugins/StaticPages/templates/default/staticpages/admin.tpl.php
@@ -34,7 +34,7 @@
 <div class="row">
     <div class="col-md-10 col-md-offset-1">
         <p class="pages">
-            <a href="<?= \Idno\Core\Idno::site()->config()->getDisplayURL() ?>staticpages/edit/?category=<?= urlencode($category) ?>"
+            <a href="<?= \Idno\Core\Idno::site()->config()->getDisplayURL() ?>staticpages/edit/"
                class="btn btn-primary btn-add">Add new page</a>
         </p>
     </div>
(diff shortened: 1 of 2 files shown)


IdnoPlugins/StaticPages/templates/default/staticpages/admin.tpl.php
```diff
@@ -34,7 +34,7 @@
 <div class="row">
     <div class="col-md-10 col-md-offset-1">
         <p class="pages">
-            <a href="<?= \Idno\Core\Idno::site()->config()->getDisplayURL() ?>staticpages/edit/?category=<?= urlencode($category) ?>"
+            <a href="<?= \Idno\Core\Idno::site()->config()->getDisplayURL() ?>staticpages/edit/"
```
kylewm: is this related?

physcocode: I was gonna ask the same

mapkyca: No, but it was causing a notice error (and therefore an exception, since I run with pedantic mode on). Needed to tidy it up first, and didn't see much point in making this a second PR

kylewm: 👍 just wanted to make sure it was on purpose

rosetree:
Please excuse my late reply. I can confirm that the commit cf3bec8269e5efa6313deb128c23fe34a467737d resolves my issue #1477. Thank you, @mapkyca!

mapkyca:
HTH :)

Issue #1031, opened 2015-08-27
working with nginx


jpcaruana:
I would like to be able to put an nginx instead of apache on my install, but I can't find anywhere configuration to make it work.
any idea ?

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/1031" class="u-in-reply-to"></a>A few folk use nginx, try asking on IRC or in the forum for a sample config :)

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/1031" class="u-in-reply-to"></a>This might be a good starting point: https://gist.github.com/kylewm/7d8cbc9a72984d67c646

jpcaruana:
thanks I'll try this !

paulcmal:
> This might be a good starting point: https://gist.github.com/kylewm/7d8cbc9a72984d67c646

This config could allow to [parse any file](https://nealpoole.com/blog/2011/04/setting-up-php-fastcgi-and-nginx-dont-trust-the-tutorials-check-your-configuration/) called with a request ending in ".php" as a PHP script, which is pretty bad.

For instance, someone uploading a picture/avatar containing embedded PHP code could then, given the address of the image, run the code on your server without you knowing.

Adding `try_files $uri =404;` to the location php block should prevent this vulnerability, as it enforces a check that the file requested actually exists and doesn't just "match the location".

To be fair, I think Known's rewrite rules practically make this exploit unusable. But it never hurts to take good/safe habits ;-)

mapkyca:
<a href="" class="u-in-reply-to"></a><a href="https://github.com/idno/Known/issues/1031#issuecomment-135958676" class="u-in-reply-to"></a>Thanks for the tip!

As an aside, once folk have a good nginx config, it'd be real handy if someone submits a pull request. I think it'd be handy to have a config example in the repo...

kylewm:
Thanks for the heads up @paulcmal! I added `try_files $uri =404;` to the gist, to hopefully avoid propagating bad configurations.

buzzworkers:
still looking for the perfect nginx conf to make Known run on a sub directory. 
I should be able to provide a working nginx for Known when installed at the root of the server.

jpcaruana:
[...]

PR #1497, opened 2016-08-11
Performant nginx configuration

pierreozoux:
https://developers.google.com/speed/pagespeed/insights/?url=http%3A%2F%2Fblog.indie.host%2F

From https://github.com/indiehosters/known/blob/master/nginx.conf

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -52,6 +52,7 @@ http {
     }
 
     location ~ \.php$ {
+      try_files $uri =404;
       fastcgi_param  GATEWAY_INTERFACE  CGI/1.1;
       fastcgi_param  SERVER_SOFTWARE    nginx;
       fastcgi_param  QUERY_STRING       $query_string;
--- a/nginx.conf
+++ b/nginx.conf
@@ -42,8 +42,6 @@ http {
       expires           30d;
       add_header Pragma public;
       add_header Cache-Control "public";
-      rewrite ^/css/[0-9]+/(.*) /css/$1 last;
-      rewrite ^/js/[0-9]+/(.*) /js/$1 last;
       try_files $uri $uri/ /index.php?$query_string;
     }
 
--- a/nginx.conf
+++ b/nginx.conf
@@ -1,27 +1,87 @@
-# Experimental nginx configuration file for domain and subdomain Known installations
-# Originally written by Kyle Mahan
+user www-data;
 
-server {
+events {
+  worker_connections 2048;
+  multi_accept       on;
+  use                epoll;
+}
 
+http {
+  sendfile           on;
+  tcp_nopush         on;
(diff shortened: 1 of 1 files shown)


nginx.conf
```diff
@@ -1,27 +1,87 @@
...
+    open_file_cache_valid    60s;
+    open_file_cache_min_uses 5;
+    open_file_cache_errors   off;
+    
+    location ~* \.(css|js|jpg|jpeg|gif|png|ico|xml)$ {
+      expires           30d;
+      add_header Pragma public;
+      add_header Cache-Control "public";
+      rewrite ^/css/[0-9]+/(.*) /css/$1 last;
+      rewrite ^/js/[0-9]+/(.*) /js/$1 last;
```
kylewm: I don't think we need these rewrites anymore

nginx.conf
```diff
@@ -1,27 +1,87 @@
...
+      try_files $uri $uri/ /index.php?$query_string;
     }
 
     location ~ \.php$ {
-        try_files $uri =404;
-        include /etc/nginx/fastcgi_params;
-        fastcgi_pass unix:/var/run/php5-fpm.sock;
-        fastcgi_index index.php;
-        fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
+      fastcgi_param  GATEWAY_INTERFACE  CGI/1.1;
```
kylewm: @paulcmal suggested `try_files $uri =404;` here: https://github.com/idno/Known/issues/1031#issuecomment-135958676

Worth keeping it?

nginx.conf
pierreozoux: in 0.9.2 I do need them. Do you mean now, in the current version we don't need them anymore?

nginx.conf
pierreozoux: You mean for the line 47 and 51?

I just tried, but it is giving me a 404...

kylewm: In `location ~ \.php$ { ... }` line 55, similar to how it was in the old file.

nginx.conf
kylewm: Ah ok, yeah the code in master doesn't need them, but I guess there hasn't been a release since then.

nginx.conf
pierreozoux: ah ok, just updated it.

nginx.conf
pierreozoux: ok, removed.

rikmendes:
Wondering if any of you could also produce a working nginx conf for Known when installed into a sub directory like yourdomain.net/blog/ up to now all configurations I have seen so far are for root installations. 

ps: I tried all the tweaks I could find last weekend to get it running on a subdir without any success.

pierreozoux:
\o/

PR #1507, opened 2016-09-08
Make PubSub functions useable by child classes

mapkyca:

Change private methods to protected

## Here's why I did it:

This is good practice, and necessary as I wanted to implement my own pubsub functions and not rewrite the wheel.


mapkyca:
Minor change, but high priority, so merging myself (cc @benwerd )

Issue #1505, opened 2016-09-08
Micropub: HTTP "authorization" header not accepted (case-sensitive)


cweiske:
## While trying to do this:
Send a micropub request with the bearer token in an `authorization` header (lowercase)

## I encountered this error:
Not authenticated

## Some other notes:
Sending the bearer token as `Authorization: Bearer 12345` works, but `authorization: Bearer 12345` (lowercase or otherwise cased) does not work.

[HTTP 1.1 RFC 2616 section 4.2](https://tools.ietf.org/html/rfc2616#section-4.2) says:

> Field names are case-insensitive.

----
Related to snarfed/wordpress-micropub#34

cweiske:
Not using the native `getallheaders()` function (as suggested in #1509) will solve this issue.

Issue #1509, opened 2016-09-08
Micropub authorization does not work on fastcgi servers


cweiske:
## While trying to do this:

Post a note via micropub on a apache web server that uses fastcgi to speak with PHP
## I encountered this error:

`403 Access denied`
## Some other notes:

Micropub bearer tokens are transmitted via the `Authorization` HTTP header.
### Authorization header

On fastcgi, the HTTP `Authorization` header is not available in `$_SERVER` (ant neither in `getallheaders()`). There is no trace of it visible.

To make pass the header to PHP, `.htacess` must be modified:

```
RewriteCond %{HTTP:Authorization} ^(.+)$
RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
```

or

`SetEnvIf Authorization "(.*)" HTTP_AUTHORIZATION=$1`

Both variants make the header available at `REDIRECT_HTTP_AUTHORIZATION`.
There is no way to get rid of the `REDIRECT_`.
### known

Known's `Idno/Common/Page::getallheaders()` uses PHP's in-built `getallheaders()` function, which does not contain the rewritten `REDIRECT_HTTP_AUTHORIZATION` header.

To fix the issue, the call to `getallheaders` must be removed and only the fallback variant be used.

The nginx fallback variant must be extendet:

```
foreach ($_SERVER as $name => $value) {
    if (substr($name, 0, 14) == 'REDIRECT_HTTP_') {
        $name = substr($name, 9);
    }
    if (substr($name, 0, 5) == 'HTTP_') {
```

With those changes (.htaccess and Page.php) micropub works.

mapkyca:
Can close?

Your patch seems to have been merged...

PR #1511, opened 2016-09-09
Make micropub authentication work on FastCGI servers

cweiske:
See full description in #1509.

When Apache connects to PHP via FastCGI, the HTTP "Authorization" header
is not sent to PHP.
.htaccess is modified so that the header is made available to PHP.

The header is available as REDIRECT_HTTP_AUTHORIZATION, so the
Idno/Common/Page::getallheaders() is adjusted to take that into account.
Usage of the native `getallheaders()` function is dropped since it does not
return the REDIRECT_HTTP_AUTHORIZATION header at all.

Resolves: #1505, #1509

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -4,6 +4,9 @@
     RewriteEngine on
     #RewriteBase /
 
+    RewriteCond %{HTTP:Authorization} ^(.+)$
+    RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
+
     RewriteCond %{ENV:BASE} ^$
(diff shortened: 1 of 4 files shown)


benwerd:
Has this been tested on non-FastCGI environments?

cweiske:
The patch is running locally here with apache2+mod_php, and on my live server with apache+fastcgi

cweiske:
What can I do to get this patch merged?

physcocode:
@cweiske I conform it working it will be merged soon as @benwerd is busy with his trip to England his home :p

PR #1520, opened 2016-09-16
Pubsub: findFeed now supports atom alternatives

mapkyca:

Added atom feed to findFeed regex

## Here's why I did it:

We were previously only detecting atom feeds.


benwerd:
Boom. Thank you.

Issue #1523, opened 2016-09-21
HTML entities in content are double-decoded


cweiske:
## While trying to do this:
Create a post with content `<this is a test>`

## I encountered this error:
Text not visible in created post

## Some other notes:
`Idno/Core/Template.php#parseHashTags()` calls `html_entity_decode` which it should not do.
That was introduced with e79211d7a24b68f90b96bdeefbd1afe2157862d5

When fixing this is must be made sure that all manual `&`-entities are kept:

`<h1>test</h1> &quot; &amp; &lt; &gt;` and so on must be rendered correctly.

PR #1526, opened 2016-09-21
Fix HTML entity decoding while keeping support for unicode tags.

cweiske:
Decode `&auml;` to "ä", but keep "<", ">" and "&" characters encoded.

Resolves: #1523


physcocode:
+1 completely logical !

cweiske:
oh. wrong merge.

benwerd:
Oh, neat. 👍

Issue #1513, opened 2016-09-12
Micropub: syndication[] parameter dropped


cweiske:
## While trying to do this:
Create a note with two "syndication" URLs on the micropub endpoint, using the `syndication` parameter.
```
$ curl -i -H 'Authorization: Bearer abc' -d 'h=entry' -d 'content=syndicatio test'\
     -d 'syndication[]=http://example.org/1' -d 'syndication[]=http://example.org/2'\
     http://example.org/micropub
```
## I encountered this error:
Syndication URLs are not visible

## Some other notes:
Using `syndication` as parameter name works, but `syndication[]` not.

The [micropub draft spec section 3.3 create](http://micropub.net/draft/#create) says:

> To specify multiple values for a property, such as multiple categories of an h-entry, use array bracket notation for the property name.

----

Related to #1508, #1512.

PR #1528, opened 2016-09-21
Support multiple syndication URLs in micropub requests

cweiske:
Resolves: #1513


physcocode:
+1 I noticed this back in time but ignored it thanks for bringing it back and offering a fix :)

Issue #1374, opened 2016-03-20
Feature Request: Add support for video uploads


neilr8133:
Feature request: Known already supports audio uploads, would love it if it could properly implement video uploads as well (cross-ref: #1373: video uploading works in some situations but seems to not be intended/reliable behavior).

I looked at the community plugin "Video" by Tyler Gillies but it is a stub and has no video-related functionality I could detect.

Thanks!  :-)

benwerd:
Video uploads would be awesome.

There's a difficult user education problem here. Videos uploaded directly
from a phone are unsuitable for upload: the bitrate, and therefore the file
size, is way too high.

Typically services transcode these files to a lower bitrate, and optimize
the file for progressive downloading. (Not every video, including MP4 h.264
files, will stream as you play.)

Self hosters have very little hope of installing or maintaining the
infrastructure required for this themselves. You often need a combination
of ffmpeg, Handbrake, and sometimes ffmbc depending on the incoming file,
and the resulting server load would force shared hosting users to bump out
of their plans.

So: either users do home transcoding themselves, which is typically a very
fiddly process, or they need to depend on third party infrastructure to do
this.

It's possibly something Convoy could do, but because of the server loads
involved, it would likely need to cost more than $5 a month for many users.
Not only that, but YouTube keeps a copy of every original file as it stood
before transcoding for legal compliance reasons; Convoy would probably need
to do that too.

So, in other words, it's hard, and is the kind of thing that centralized
services can do far more easily. I don't write this expecting to
necessarily figure out a solution - but I figured it would be helpful to go
into the reasons why it's not there.

On Sunday, March 20, 2016, neilr8133 <notifications@github.com> wrote:

> Feature request: Known already supports audio uploads, would love it if it
> could properly implement video uploads as well (cross-ref: #1373
> <https://github.com/idno/Known/issues/1373>: video uploading works in
> some situations but seems to not be intended/reliable behavior).
>
> I looked at the community plugin "Video" by Tyler Gillies but it is a stub
> and has no video-related functionality I could detect.
>
> Thanks! :-)
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/idno/Known/issues/1374>
>


-- 
Ben Werdmuller
<http://goog_1933028737>
benwerd.com | werd.io

+1 (312) 488-9373

neilr8133:
Thank you for the excellent summary.

[...]

PR #1531, opened 2016-09-22
Add video upload support to the micropub endpoint

cweiske:
Requires a video plugin to be installed, e.g.
https://github.com/cweiske/withknownVideo

Related: #1374

### DIFF
--- a/IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
+++ b/IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
@@ -137,7 +137,7 @@
                         $type = 'photo';
                     } else if ($photo_url = $this->getInput('photo')) {
                         $type      = 'photo';
-                        $success   = $this->uploadFromUrl($photo_url);
+                        $success   = $this->uploadFromUrl('photo', $photo_url);
                         if (!$success) {
                             \Idno\Core\Idno::site()->triggerEvent('indiepub/post/failure', ['page' => $this]);
                             $this->error(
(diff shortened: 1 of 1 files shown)


IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
```diff
@@ -111,14 +111,29 @@ function post()
...
                         if (!$success) {
                             \Idno\Core\Idno::site()->triggerEvent('indiepub/post/failure', ['page' => $this]);
                             $this->setResponse(500);
                             echo "Failed uploading photo from $photo_url";
                             exit;
                         }
-                    } else if (!empty($name)) {
+                    }
+
+                    if (!empty($_FILES['video'])) {
```
kylewm: `else if`? I'm assuming a post cannot have both a photo and a video

IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
```diff
@@ -111,14 +111,29 @@ function post()
...
+                        $success   = $this->uploadFromUrl('video', $video_url);
+                        if (!$success) {
+                            \Idno\Core\Idno::site()->triggerEvent('indiepub/post/failure', ['page' => $this]);
+                            $this->setResponse(500);
+                            echo "Failed uploading video from $video_url";
+                            exit;
+                        }
+                    }
+
+                    if ($type == 'note' && !empty($name)) {
```
kylewm: If you use `else if` here too, you can remove the `$type == note` check

IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
cweiske: Some micropub clients (ownyougram and shpub) upload a video and the preview image at the same time, so having both is actually used. Known throws the photo away right now, but that can be changed in the future.

IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
cweiske: no, I can't because of above.

benwerd:
This, again, looks like it could be really cool (although it would be good to have a POC video plugin that supports both images and video). I'm in favor of merging once conflicts have been resolved.

High five @cweiske for contributing to core!

cweiske:
So what can I do to get this patch merged?

physcocode:
@kylewm @cweiske I think this is ready to merge I see no conflict here . Ownyourgram actually uploads both image and preview at same time !

kylewm:
Sorry for the delay; I didn't see that you'd fixed the merge conflicts :+1:

Issue #1506, opened 2016-09-08
Micropub: Error responses are not JSON


cweiske:
## While trying to do this:
Send an invalid micropub request to create a new post

## I encountered this error:
Server responded with a status code of 500 and plain-text error message "Couldn't create note"

## Some other notes:
The [micropub draft specification section 3.8 Error Response](https://www.w3.org/TR/micropub/#error-response) states:

> If an error body is returned, the response body must be encoded as a [JSON] object and include at least a single property named error with the value set to one of the below:

Such an error can be forced by making the `content` POST parameter an array:

    $ curl -X POST -H 'Authorization: Bearer foo' -d h=entry -d 'content[]=test' http://known.example.org/micropub/endpoint
    HTTP/1.0 500 Internal Server Error
    ...
    Content-Type: text/html
    
    Couldn't create note

cweiske:
Same error occurs when the access token is invalid

PR #1532, opened 2016-09-22
Send proper micropub error responses

cweiske:
Conform to the spec by sending out JSON error responses:

- When the access token is invalid
- When some error occurs, e.g. a post could not be saved

The micropub endpoint now also sends out a HTTP 500 status code
by default unless that's overriden later.
This helps when something crashes during processing.

Resolves: #1506

### DIFF
--- a/IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
+++ b/IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
@@ -80,6 +80,9 @@
 
             function post()
             {
+                //fail-by-default in case of unhandled errors
+                $this->setResponse(500);
+
                 $this->gatekeeper();
                 // If we're here, we're authorized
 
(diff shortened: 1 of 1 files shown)


IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
```diff
@@ -227,20 +231,58 @@ function post()
...
                 }
+
+                $this->setResponse($statusCode);
+                header('Content-type: application/json');
+                echo json_encode(
+                    array(
+                        'error'             => $error,
+                        'error_description' => $description,
+                    )
+                ) . "\n";
```
kylewm: Maybe use Known's JSON template since it'll set the proper headers and whatnot.

e.g.,

```
                Idno::site()->template()->setTemplateType('json');
                Idno::site()->template()->__([
                    'notifications' => $arr,
                ])->drawPage();
```

benwerd:
High five to @kylewm for reviewing these!

IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
benwerd: Seconded! Please use our JSON template output. If that isn't sufficient for whatever reason, it'd be good to expand the template to support whatever you need.

cweiske: I tried that; it works but adds the `messages` key - unconditionally. This isn't against the micropub spec since that one does not have a schema for errors anyway.

cweiske:
Ok, switched to the template.

Issue #1518, opened 2016-09-15
Micropub: "bookmark-of" property ignored


cweiske:
## While trying to do this:
Bookmark an URL via micropub

```
curl -X POST -H 'Content-Type: application/x-www-form-urlencoded'\
    -H 'Authorization: Bearer abc'\
    -d 'h=entry' -d 'bookmark-of=http://www.bogo' -d 'content=foo'\
    'http://known.bogo/micropub/endpoint'
```

## I encountered this error:
Normal post is created instead of a bookmark

The bookmark plugin is enabled.

PR #1533, opened 2016-09-22
Add support for bookmarks in micropub

cweiske:
Resolves: #1518


kylewm:
I don't think I know enough to review this one -- I always had trouble with the Like plugin since it can't make up its mind whether it wants to be "likes" or "bookmarks"

cweiske:
Oh, it's easy: Just use shpub to create a bookmark and a like:

```
$ ./bin/shpub.php like http://example.org/
Like created at server
http://known.bogo/2016/example-domain

$ ./bin/shpub.php bookmark http://example.org/
Bookmark created at server
http://known.bogo/2016/example-domain-1
```

You'll see the difference.

PR #1535, opened 2016-09-22
Support annotation creation via micropub (like/reply(comment))

cweiske:
Ability to create comments and likes as annotation via micropub

## Here's why I did it:
I want to back up instagram posts together with their likes and their comments, so I needed a way to import them programmatically.

### DIFF
--- a/IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
+++ b/IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php
@@ -87,6 +87,10 @@
 
                 // Get details
                 $type        = $this->getInput('h', 'entry');
+                if ($type == 'annotation') {
+                    return $this->postCreateAnnotation();
+                }
+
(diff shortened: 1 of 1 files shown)


kylewm:
Interesting, this is clever extension of micropub, but couldn't it be done more easily by webmention/bridgy?

physcocode:
agreed with @kylewm but this should definitely merged into core !

cweiske:
@kylewm: bridgy would need to get (private) instagram support, which the instagram API does not allow for not-own-private-images. 

If you want to have that with webmention only, you'd need to setup fake comment html with microformats. Just having a method to push comments and likes into the system is easier.

benwerd:
I think this is really cool :) We can experiment with it - I think it's worth having it in core. Webmention is also a thing, but direct comments are neat.

NB: I do, however, disagree with syndicating comments from private spaces elsewhere, unless everyone in that space has consented to the syndication.

PR #1541, opened 2016-09-27
Remove "This is an IndieAuth server" message.

cweiske:
This is not needed anymore since the bug in IndieAuth.com is fixed
that required this message.

See https://github.com/aaronpk/IndieAuth.com/issues/93

Follow up for #1537


kylewm:
heh, so it's OK with HTTP 403 now so long as it has the header?

cweiske:
Yes, it is now - https://github.com/aaronpk/IndieAuth.com/commit/69048d6bbb2bcd1dbeb0f0886b56306a789a971c

PR #1544, opened 2016-09-30
Admin acl tests

mapkyca:

Some tests for ACLs, some fixes for logged out users

## Here's why I did it:

ACLs weren't being tested, now they are.

Closes #1543

### DIFF
--- a/Idno/Common/Entity.php
+++ b/Idno/Common/Entity.php
@@ -755,7 +755,8 @@
                         $this->deleteData();
                         \Idno\Core\Idno::site()->triggerEvent('deleted', array('object' => $this));
 
-                      if(!empty($this->getAttachments()))
+                      $attachments = $this->getAttachments();
+                      if(!empty($attachments))
                       {
                         $this->deleteAttachments();
                       }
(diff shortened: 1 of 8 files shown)


mapkyca:
Hold off from merging just yet, I need to isolate some slug generation errors this is causing...

benwerd:
@mapkyca Yep, way ahead of you :) Let's figure this out before it goes in. Thanks for writing tests though!

mapkyca:
It is interesting... I think these issues aren't to do with the access tests per se, rather they've exposed a bug with object construction. Not sure what's going on, but it seems that getByID is returning an object with an incorrect UUID.

mapkyca:
Ok, zeroing in... looks like there's a double save error going on, possibly with test, possibly with the generic data item.

An initial object is created in the database with a generic UUID, and _ID. Then a second object is saved with a corrected stub based UUID and a new _ID, with the correct UUID and the old _ID value present in the json blob.

This causes the value the constructed entities to look the same, but the getByID and getByUUID methods to return different values. This seems to be a new issue...

mapkyca:
Further... this is damn weird. Save is only being called once, however it definitely seems to be inserting two records...

mapkyca:
Ahh... it's a db state issue, combined with some other weirdness.

mapkyca:
Also fixed a PHP 5.4 error for free.

Issue #1277, opened 2016-02-20
Drop Mongo support


mapkyca:
Kill it. Kill it with fire.

Refs #1271
Closes #1241 
Closes #1018 

Depends on #480

physcocode:
why are we still supporting it ? any special reason ?

mapkyca:
We need to support it until import/export is bullet proof, as we need to provide a way to migrate from mongo to mysql.

physcocode:
+1 sounds cool

mapkyca:
Wondering whether this is still necessary, given that #1241 removes the PHP7 blocker, and that indexes remove the pagination errors.

Still deprecated, but...

mapkyca:
Going to close this. The recent Mongo code migration has resolved many of the outstanding mongo issues and all the PHP7 blockers, however mongo is still deprecated, so use mysql.

PR #1551, opened 2016-10-13
Migrate away from legacy Mongo interface to MongoDB

### LINKED ISSUE #787: Unify File interfaces
If you're using Mongo, files are returned as MongoGridFSFile, otherwise they're Idno\Entities\File.

This presents a problem, since operations on those files return objects with different interfaces, which complicates logic whenever you need to query the datastore object.

The mongo datastore should probably encapsulate the MongoGridFSFile automatically within a child of Idno\Entities\File, and present a similar interface for attributes and methods.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9478334-unify-file-interfaces?utm_campaign=plugin&utm_content=tracker%2F300028&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F300028&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1241: Migrate away from mongo to mongodb
The mongo extension is legacy (https://secure.php.net/mongo), and has been replaced by the newer mongodb driver (https://secure.php.net/mongodb).

Support for mongo has (apparently) been dropped in PHP 7, so we need to migrate. 

Refs #1239 

This library (https://github.com/mongodb/mongo-php-library) might be useful, as it contains a similar interface as the legacy driver.

mapkyca:

Migrated mongo support from the legacy mongo driver to the newer MongoDB driver.

## Here's why I did it:

mongo is deprecated and no longer supported in PHP 7. While we definitely want to retire mongo, due to other issues, this will give those running mongo a little runway.

IMPORTANT: This is a major change. Back up and kick the tires before you change your mongo DB engine.

Closes #1241 
Closes #787
Refs #1277


benwerd:
Eep. I'll leave it to you to correct the merge conflict. LG after that

PR #1553, opened 2016-10-20
Fixing notice errors

mapkyca:

Fixed various notice errors.

## Here's why I did it:

Various notice errors were found, which broke stuff in pedantic mode


physcocode:
The draw back of converting notice to exception :p

mapkyca:
Not a drawback, this was the desired effect... makes notice errors impossible to ignore and enforces best practice!

physcocode:
@mapkyca totally agree and I actually totally support adding this to core than pediatic mode!

PR #1555, opened 2016-10-20
Fix build errors introduced by a672895d388a1fcecc8f30565d57044e1a9b9e43.

mapkyca:

Specify and activate some default plugins

## Here's why I did it:

Allow the tests to pass


mapkyca:
Going to go ahead and merge this down, as the current build errors prevent anyone else's pull requests from being checked properly. 

CC/ @benwerd @erinjo

PR #1556, opened 2016-10-20
Symfony update

mapkyca:
- Updated symfony components: ClassLoader, Console, EventDispatcher, and HttpFoundation to latest github
- Introduced an interface shim for UniversalClassLoader
## Here's why I did it:

The components were old, and for one didn't support the new Mongo interface (see #1241)


mapkyca:
Looks like this isn't a simple upgrade... the new PDO and PDO Session handlers appear to be incompatible with how we've currently implemented database transactions.

Probably means an error on our side... will have a play when I have more time

mapkyca:
Still scratching my head over this one... runs locally (apache + mod php), but gives an error on travis (fast-cgi, I guess)

Anyone running fast cgi want to tell me what the error is?

mapkyca:
Going to kill this... a symfony upgrade at this stage looks more involved than I'd like to pick at. Pull requests welcome, and folk are welcome to use this branch as a base, but I'm not going to work on this further unless there's a real need (beyond having mongo sessions, which I'm pretty sure was a bad idea anyway)

PR #1560, opened 2016-10-21
Introducing inACL() helper function

mapkyca:

Added a inACL() helper function

## Here's why I did it:

Needed a method to determine whether you had access to something because you had access to something or because you were an admin.


mapkyca:
Going to merge this down, since it's required for some project work.

PR #1562, opened 2016-10-21
Revert "Adding admin bypass to canRead"

erinjo:
Reverts idno/Known#1552

### DIFF
--- a/Idno/Common/Entity.php
+++ b/Idno/Common/Entity.php
@@ -1293,12 +1293,6 @@
                 if ($access == 'PUBLIC') return true;
                 if ($access == 'SITE' && \Idno\Core\Idno::site()->session()->isLoggedIn()) return true;
                 if ($this->getOwnerID() == $user_id) return true;
-                
-                if ($user = User::getByUUID($user_id)) {
-                    if ($user->isAdmin()) {
-                        return true;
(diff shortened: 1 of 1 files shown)


physcocode:
@erinjo @mapkyca any particular reason for reverting ?

erinjo:
@physcocode Hi Pranjal. Yes. We're working through some access permission updates. I reverted this because it allows site administrators to see other users' private posts.

mapkyca:
So, the admin user as a special user is a pretty standard model for multiuser platforms (wordpress, elgg, etc), and is necessary for e.g. handling abusive posts, hellbanning, etc etc. 

It was also definitely how Known was _intended_ to work, see admin bypasses elsewhere in canEdit() etc, so a bypass on canRead() is definitely how it was _intended_ to work.

If by private posts you're really talking about drafts, and the issue where draft posts were appearing in a feed for admin users, then this will largely be resolved by #1559, meaning drafts can be hidden via a filter, and avoids using ACLs for this purpose.

physcocode:
@mapkyca I was actually thinking same Admin needs to be "Admin" and canEdit goes hand to hand with canRead

mapkyca:
Closing, no longer relevant.

PR #1569, opened 2016-10-25
Provide a backup cache directory for HTML purifier

mapkyca:

If upload_dir isn't specified, purifier will use system temp for cache data

## Here's why I did it:

To remove this errrors.


mapkyca:
Merging this down as it's a minor, non-visible change

PR #1570, opened 2016-10-25
Unterminated shell

mapkyca:

* Handle cases when an object was created by a deleted user (rare, but happened on my test build owing to broken phpunit tests)
* Encode entity title to allow for & and similar characters.

## Here's why I did it:

RSS was broken and spat out errors


mapkyca:
Merging down, since it's a minor non visible change.

PR #1577, opened 2016-10-29
Subdirectory menus

mapkyca:
- Introduced doesPathMatch() helper function
- Modified admin and account menus to use it
## Here's why I did it:

Current menu highlighting did not work in subdirectory installs.


mapkyca:
May have a visual impact, so letting @benwerd merge

mapkyca:
Actually, thinking about it, this is actually a bugfix which shouldn't have a visual impact, so merging.

PR #1578, opened 2016-10-29
Deprecate PHP 5.5 support

mapkyca:
- Diagnostics and warmup will nag on both PHP 5.4 and PHP 5.5
- Installation requirements now documented as PHP 5.6+
- Moved 5.5 build to travis' allowed_failures branch
## Here's why I did it:

PHP 5.5 is dead, and will not be receiving security patches across the board, we should no longer go out of our way to support it. 

As with previous patch for 5.4, this isn't a hard failure, so those stuck on 5.5 can still install (with a massive warning), but the travis change means that while we will make best efforts to maintain support, we won't be held back.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ env:
 matrix:
     allow_failures:
         - php: 5.4
+        - php: 5.5
 
 branches:
   only:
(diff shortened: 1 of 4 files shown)


mapkyca:
This sort of change has been controversial in the past, so letting @benwerd merge

benwerd:
I'm much more inclined to support this these days.

mapkyca:
Heh! good to know ;)

PR #1580, opened 2016-10-29
Make parseUsers() pluggable for in_reply_to events

mapkyca:

Trigger an event for default in_reply_to parseUsers() callback

## Here's why I did it:

So plugins can extend @whatever users


mapkyca:
Merging since it has no visible change.

PR #1581, opened 2016-10-29
Manual address entry

mapkyca:
- Fix initMap so that it sets hidden values on init.
- Update map when the address field is changed, allowing manual address input and lookup
## Here's why I did it:

Location API is restricted to TLS origins, and we currently have no fallback for those running HTTP


mapkyca:
Leaving for @benwerd to review

mapkyca:
Bugfix, so shouldn't have too great a visual impact, especially as it's isolated to a plugin, merging...

PR #1617, opened 2016-12-19
Console debug

mapkyca:

* Gave the Known Console a bit of a polish
* Made known.php load a minimal Known core boot
* Introduced the ConsolePlugin class to simplify console plugin writes
* Made known.php use class loaders to load console plugins
* Added an example plugin, and a basic debug plugin

## Here's why I did it:

I needed to start building out some debug tools to help with development


physcocode:
This is defiantly what i wanna see merging in core then maybe i can make a console plugin to build a bare plugin for known ! something like what artisan in laravel does !

mapkyca:
I've moved the potentially risky plugin to its own dev repo, so merging in the core, which I think is fairly safe (since it was broken before, so I figure nobody was actually using it!)

PR #1623, opened 2016-12-30
isAcceptedContentType() respects q parameter (new default behaviour)

mapkyca:

By default, isAcceptedContentType() now respects the q parameter

## Here's why I did it:

application/json anywhere in the accept string would override any other content type. This update changes the default behaviour to only return true if $contentType was the first/most high priority value stated.

### DIFF
--- a/Idno/Common/Page.php
+++ b/Idno/Common/Page.php
@@ -853,14 +853,39 @@
             /**
              * Detects whether the current web browser accepts the given content type.
              * @param string $contentType The MIME content type.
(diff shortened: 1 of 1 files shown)


mapkyca:
@rhiaro @snarfed, let me know if this resolves your issue.

/cc @benwerd @erinjo

snarfed:
awesome! i don't have a known instance set up to test on, and i don't know if @rhiaro does, so we may take your word for it. thank you!

PR #1627, opened 2017-01-10
Only load initial_plugins the once

mapkyca:

Unset initial_plugins if present in config, as loaded from the database.

## Here's why I did it:

Not doing this meant that if a plugin was present in the ini but not in the database then the merged array contained the missing setting.

This is ONE PROPOSED solution, another would be to reverse the precedence so that database settings ALWAYS replace config settings, but that might be undesirable in other ways (for example, it's quite often desirable to have ini trump database for many things)

### DIFF
--- a/Idno/Core/Config.php
+++ b/Idno/Core/Config.php
@@ -194,6 +194,10 @@
                 }
 
                 if (!empty($this->ini_config)) {
+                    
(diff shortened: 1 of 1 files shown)


mapkyca:
It's worth noting that I've been up working since 4am, so this may not be the best solution on those grounds alone...

PR #1643, opened 2017-01-30
Auto archive to wayback machine

mapkyca:

When a URL is archived, by default, also post it to the wayback machine, so there's a historic record.

## Here's why I did it:

Trump


mapkyca:
Also see my plugin, which is an alternative (better) way of doing this..?

benwerd:
Yeah, let's stick with an external plugin.

PR #1645, opened 2017-02-08
Replace TinyMCE with QuillJS

mapkyca:

s/tinymce/quilljs/g

## Here's why I did it:

Folk were having various problems with the tinymce editor (pasting, images etc etc) so try quill.

Notes: Quill works... differently... and so I had to do what I think is a fairly nasty hack to get it to work with <textarea> and form submission. Seems to work, but needs to be kicked about.


mapkyca:
I'm requesting a review of this since I can't help feeling that the textarea update is a *horrible* hack. Seems to work, but I can't help feeling that this is going to cause problems... want a second opinion.

erinjo:
@mapkyca Thanks. We'll be reviewing later today.

mapkyca:
Merged down, since the "hack" seems to be "the way", and I need to expand the testing...

PR #1661, opened 2017-03-09
Few more FA icons

srcr:
extended the FA icon set a little for a few common sites.

Added resolving the Font Awesome icons for reddit, instagram and paypal.

## Here's why I did it:
It's a small fix but I believe it improves the ecstatic of the profile page.


srcr:
Awesome, Thank you mister Povey!

Issue #1664, opened 2017-03-17
Micropub: audio upload not supported


martymcguire:
## While trying to do this:

Making an audio post via micropub with [Screech](https://screech.schmarty.net/).

## I encountered this error:

The audio does not appear in the contents of the post. It looks like the uploaded file is dropped altogether.

## Some other notes:

It looks like the `master` version of the IndiePub plugin now supports photo and video uploads, but audio is ignored:

https://github.com/idno/Known/blob/v0.9.2/IdnoPlugins/IndiePub/Pages/MicroPub/Endpoint.php#L89

mapkyca:
Closed by #1665

PR #1665, opened 2017-03-17
Support for audio upload via micropub

martymcguire:

Added support for `audio` in posts submitted via micropub by:

* adding `'audio'` as an alias for the `media` content type from `IdnoPlugins\\Media\\Media`
* in `Micropub/Endpoint.php`, detecting `$_FILES['audio']` and:
    * setting the post type to `audio`
    * setting `$_FILES['media'] = $_FILES['audio']` (`IdnoPlugins\\Media\\Media` is hardcoded to look in `$_FILES['media']`.
        * this raises the question of whether posting videos via micropub works, #1531 uses `'video'` type, without making it an alias of media and without aliasing `$_FILES['video']` to `$_FILES['media']` for the Media plugin to pick it up.

This has been tested against a fresh install of Known running on the `master` branch.

## Here's why I did it:

* Known supports Audio files out of the box via the Media plugin,
* [`audio` is an acceptable micropub file upload type](https://www.w3.org/TR/micropub/#posting-files) and
* I am the developer of [Screech, a micropub client for posting audio](https://github.com/martymcguire/screech), which posts audio files as `$_FILES['audio']`.


martymcguire:
Note that this closes #1664.

martymcguire:
@aaronpk added support for audio URLs in JSON-encoded micropub requests.

during development, we each noticed that video uploading via micropub probably wasn't working. @aaronpk confirmed this and a fix is included. he also made a couple of test cases for micropub video uploads which can be found here:

https://gist.github.com/aaronpk/f8ed77c8537369cce8eee959ef4aa205

@aaronpk also encountered a strange issue when testing video upload via micropub. A new post is created successfully, the video file is saved to the server, and the post displays with a `<video>` tag as expected. However, the embedded video does not play in Chrome or Firefox. Oddly, the path to the raw .mp4 also does not successfully load in Chrome or Firefox.

The dev tools network tab shows that when requesting the .mp4, a 200 response is returned with the correct content length, but only partial content. The browser follows up with partial content requests for the rest and gets back a 206 response that seems to be off-by-one. Here's an example Content-Length response header:

`Content-Range: bytes 5308416-5380899/5380900`

Interestingly, fetching the .mp4 with curl delivers the entire file:

```
$ wget https://known.schmarty.net/file/d3ac4badaad1e8ae126a0faa10a172f4/video.mp4
--2017-03-30 12:05:34--  https://known.schmarty.net/file/d3ac4badaad1e8ae126a0faa10a172f4/video.mp4
Resolving known.schmarty.net... 159.203.95.59
Connecting to known.schmarty.net|159.203.95.59|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 5380900 (5.1M) [video/mp4]
Saving to: ‘video.mp4’

video.mp4              100%[==========================>]   5.13M  3.05MB/s    in 1.7s

2017-03-30 12:05:36 (3.05 MB/s) - ‘video.mp4’ saved [5380900/5380900]
```

This issue doesn't seem related to micropub handling of video/audio and should probably be its own PR.

aaronpk:
Note that the video plays fine in Safari though!

martymcguire:
Another video playback update: uploading the video file via Known's native "audio" media uploader (which accepts video) produces the same behavior:

https://known.schmarty.net/2017/test-video-upload-w-native-known-interface

martymcguire:
Ok, I made a note of the off-by-one video serving behavior.

Since it's not related to the micropub handler specifically, I think this PR should be good to merge.

PR #1673, opened 2017-03-25
Added syndication plugin for pnut social network

chrisaldrich:
pnut.io is a Twitter-like status update social network

Also added note to plugin for App.net, which has been shut down. Please keep the link to the plugin as it was actually used as a template for the pnut plugin and may be useful to others in the future for this reason.


benwerd:
I've merged this. I think there's something to be said, though, for making a true syndication plugin template, rather than using App.net's as a basis.

PR #1675, opened 2017-03-27
Bringing back previous wrapping functions

mapkyca:

Restored previous mongo key wrapping keys

## Here's why I did it:

Mongo was broken for some people


benwerd:
This worked!

mapkyca:
Cool, merging this into mainline...

PR #1695, opened 2017-04-13
Adding Watching plugin

cleverdevil:

Added a new plugin for tracking what you watch (TV and movies), including a Plex webhook.

## Here's why I did it:

Because I wanted to!


danito:
:1

PR #1709, opened 2017-04-21
Adding the ability to query with "<" and ">" operators to MySQL.

cleverdevil:

I added the capability to query with "<" and ">" operators to the MySQL backend. If I also need to add support for other database backends to get this PR merged, please do let me know, and I'll try and work on those too. From what I could tell, the only supported backend at this point is MySQL, but I could be wrong!

## Here's why I did it:

To enable me to create an "On This Day" feature for my website. You can find my implementation here:

    https://github.com/cleverdevil/CleverCustomize/blob/master/Pages/OnThisDay.php

... you can see it in action here:

    https://cleverdevil.io/onthisday/
    https://cleverdevil.io/onthisday/04/18

### DIFF
--- a/Idno/Data/MySQL.php
+++ b/Idno/Data/MySQL.php
@@ -527,6 +527,42 @@
                                 }
                                 $subwhere[] = $instring;
                             }
+                            if (!empty($value['$<'])) {
+                                $val = $value['$<'];
+                                if (in_array($key, $this->getSchemaFields())) {
+                                    $subwhere[] = "(`{$collection}`.`{$key}` < :nonmdvalue{$non_md_variables})";
+                                    if ($key === 'created') {
+                                        if (!is_int($val)) {
+                                            $val = strtotime($val);
+                                        }
+                                    }
+                                    $variables[":nonmdvalue{$non_md_variables}"] = $val;
+                                    $non_md_variables++;
+                                } else {
+                                    $metadata_joins++;
+                                    $subwhere[]                           = "(md{$metadata_joins}.`name` = :name{$metadata_joins} and md{$metadata_joins}.`value` < :value{$metadata_joins} and md{$metadata_joins}.`collection` = '{$collection}')";
+                                    $variables[":name{$metadata_joins}"]  = $key;
+                                    $variables[":value{$metadata_joins}"] = $val;
+                                }
+                            }
+                            if (!empty($value['$>'])) {
+                                $val = $value['$>'];
+                                if (in_array($key, $this->getSchemaFields())) {
+                                    $subwhere[] = "(`{$collection}`.`{$key}` > :nonmdvalue{$non_md_variables})";
+                                    if ($key === 'created') {
+                                        if (!is_int($val)) {
+                                            $val = strtotime($val);
+                                        }
+                                    }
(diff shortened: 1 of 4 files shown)


mapkyca:
Looks reasonable, thanks!

I would like to see feature parity across engines if possible. SQLite is still rather experimental, but Postgres and Mongo installs definitely exist in the wild...

cleverdevil:
OK, I'll start with PostgreSQL, since that should be a relatively simple port. Much of the code is the same between MySQL and PostgreSQL backends.

Mongo is officially deprecated in the documentation, so I'd rather not spend the effort there unless its deemed absolutely necessary.

cleverdevil:
Okay, added PostgreSQL as well!

mapkyca:
Well, Mongo is deprecated in that we really want to discourage new folks from using it, but all early adopters will be on it (including me!), and this will be true until export/import is rock solid and nobody has really had the bandwidth to get into that...

cleverdevil:
I can definitely research and work on the MongoDB side if I have to, that said, the only downside to moving forward without it is that those on MongoDB won't be able to use plugins that use this feature. I think that's a fair tradeoff, as no one's sites will break.

That said, if its absolutely necessary, that's fine, I'll set aside some time and try and figure it out :)

cleverdevil:
Okay, updated to include SQLite support. Still haven't dived into Mongo, yet. I may find time tomorrow.

mapkyca:
Cool!

Extra credit bonus, an extension to the Data Concierge unit tests to test this path would be awesome. ;)

cleverdevil:
Okay, I tinkered with setting up a MongoDB-backed Known instance using the latest and greatest from master, and while I was able to get it running, it doesn't actually seem to work. The user I created can't even log in. Unless I'm doing something wrong, it may be broken!

So, now I am shifting my attention to updating the unit test suite, as requested, but there isn't any documentation on how to actually run the tests, as far as I can tell. Can anyone point me in the right direction?

mapkyca:
Take a look at Tests/Data, I'd add the range tests to DataConciergeTest.php. You can run tests with phpunit. You may need to export KNOWN_DOMAIN='your.local.domain' if you're running on anything other than localhost.

Troubling about Mongo... I run mongo, although I've not done a from scratch install for a while. Any errors?

mapkyca:
Just tried a from fresh mongo install with latest code and was unable to replicate - the onboarding process worked as expected and my new site and first user were created with no problems...

cleverdevil:
Awesome, thanks for that @mapkyca. I probably just screwed something up during setup process with MongoDB. I'll wipe my install and try again!

Thanks for the guidance on the tests, too. I'll give it a shot.

cleverdevil:
Okay, determined that MongoDB already supports "<" and ">" with no changes, it just spells them as `$lt` and `$gt`, so I moved to that syntax globally, and also cleaned up a few things I had wrong in PostgreSQL and Sqlite backends.

Next (and last), are unit tests.

cleverdevil:
Oh, and I found out why MongoDB wasn't working: I was missing a dependency.

cleverdevil:
Okay, added unit tests. They even pass.

Issue #1725, opened 2017-04-24
Photo plugin no longer allows for selection of photo from library on iOS


cleverdevil:
## While trying to do this:

On the latest Known from master, when attempting to create a new Photo post, tapping the "Select a photo" button on iOS in Safari brings up the camera...

## I encountered this error:

... BUT it no longer allows for selecting a photo from the library.

## Some other notes:

To fix, it appears that the markup should be changed back to:

```html
<input type="file" name="photo" id="photo" class="col-md-9 form-control" accept="image/*;capture=camera" onchange="photoPreview(this)" />
```

The key is that the `capture=camera` parameter is passed as part of the `accept` rather than being on its own as an attribute. I'll try and open a pull request for this at some point, though right now I am focused on my other open pull request.

aaronpk:
It looks to me like Known always intended to open the camera directly, which is what the `capture=camera` parameter on the `accept` attribute was for. It appears that in the HTML spec, this was changed to its own attrubute, e.g. `accept="image/*" capture="camera"`.

Personally I prefer being able to choose the camera or my photo library whenever I make photo posts, since I often take photos first with my camera, edit them in an iOS image editor and save them back to my library to upload the post-processed version.

I believe the better fix to restore this feature is to just remove the `;capture=camera` bit from the `accept` attribute.

cleverdevil:
Thanks for the clarification, @aaronpk. Yes, that makes sense.

mapkyca:
This was changed as part of 580c42d25ca841219f524d79911f5cd42ee1556f

Issue was that with the capture= in the accept, the image/* was seemingly ignored and you could upload non-image files. The separate capture attribute is w3c spec, presumably there must be a way to give the option to do both...

aaronpk:
Assuming Safari and Android are tracking the w3c spec, try setting the attribute with no value.

`<input type="file" accept="image/*" capture>`

https://w3c.github.io/html-media-capture/#the-capture-attribute

cleverdevil:
Tried it, and that goes back to the behavior of popping up the camera, so sadly that won't fix the behavior.

aaronpk:
Removing the `capture` attribute entirely should surely solve it then? That should switch it back to the default on iOS which opens the sheet letting you choose between camera and upload. I'm not sure what happens on Android though.

cleverdevil:
[...]

PR #1728, opened 2017-04-26
Issue 1725

cleverdevil:

Fixing #1725 by removing `capture` attribute.

## Here's why I did it:

On iOS, tapping "select a photo" when creating a photo post only allows taking a new photo with the camera, not selecting an existing photo.


cleverdevil:
Gah, branched from the wrong spot. Fixing...

PR #1732, opened 2017-05-03
Fixing #1700 by properly decoding entities into unicode.

### LINKED ISSUE #1700: Like of certain Twitter URL's via Quill don't properly represent "special" characters
## While trying to do this:

Using [Quill](https://quill.p3k.io), if I perform a "Like" to my Known site, which is running the latest from HEAD, of [this Twitter permalink](https://twitter.com/benwerd/status/852378346822422528):

## I encountered this error:

... [a post is successfully created](https://cleverdevil.io/2017/ben-werdmuller-on-twitter-the-indieweb-summit-is-all-about), but Known represents the apostrophe like so:

```I&#39;ll```

This happens in any "Like" or "Repost" from Quill to my Known site that contains apostrophes, newlines, or other non-ASCII characters.

## Some other notes:

```
🕷  cat version.known
version = "0.9.5"
build = 2017033101
```

cleverdevil:

Changed from using `htmlspecialchars_decode` to `html_entity_decode` for more complete processing of title text in HTML.

## Here's why I did it:

This fixes #1700

### DIFF
--- a/IdnoPlugins/Like/Like.php
+++ b/IdnoPlugins/Like/Like.php
@@ -60,7 +60,7 @@
                 $str = \Idno\Core\Webservice::file_get_contents($Url);
                 if(strlen($str) > 0){
                     if ($result = preg_match("/\<title\>(.*)\<\/title\>/siu",$str,$title)) {
-                        return htmlspecialchars_decode($title[1]);
+                        return html_entity_decode($title[1], ENT_QUOTES | ENT_XML1, 'UTF-8');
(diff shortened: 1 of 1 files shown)


cleverdevil:
Without tests to show me, I am not really sure what the desired functionality is. All I know is that before my change, my site was exhibiting the behavior described in #1700 and after it isn't. What do you think is the best path forward? Maybe @erinjo or @benwerd has input?

PR #1743, opened 2017-05-11
Tombstones full uuid

mapkyca:

Contains all patches from #1742, plus full UUID uniqueness checks on the creation of new entities.

## Here's why I did it:

Enforce FULL uuid uniqueness and tombstoning across all objects.

Closes #1742


mapkyca:
Open question, since we now fully tombstone uuids, is it necessary any longer to tombstone slugs? I can see arguments on both sides...

PR #1761, opened 2017-05-22
JSONFeed

cleverdevil:

Adding support for JSONFeed.

## Here's why I did it:

It seems to be an emerging standard with potential.

### DIFF
--- a/templates/json/shell.tpl.php
+++ b/templates/json/shell.tpl.php
@@ -70,11 +70,12 @@
                 $feedItem['author']['avatar'] = $owner->getIcon();
             }
 
-            $feedItem['content_html'] = $item->draw();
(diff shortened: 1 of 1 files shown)


mapkyca:
Thanks for the PR!

My €0.02: it's probably useful to have a more standards compliant version of json output than just dumping $vars, my only concern is that the API uses the json export quite heavily and this is likely to be a breaking change for anyone using the API - the json endpoint is not exclusively used for listing objects on a GET query.

An obvious simple solution would be to rework this PR to use templates/jsonfeed/ or similar instead of templates/json/

cleverdevil:
Makes sense, I'll rework the PR as directed!

PR #1771, opened 2017-05-31
JSON Feed evolution

cleverdevil:

Decentralizing the metadata generation for JSON Feed and delegating it to plugins, where possible. Also, updating a few pieces of terminology/properties to more closely match other Indieweb work like JF2.

## Here's why I did it:

Largely in response to ongoing discussions here:

https://github.com/brentsimmons/JSONFeed/issues/51

### DIFF
--- a/templates/jsonfeed/shell.tpl.php
+++ b/templates/jsonfeed/shell.tpl.php
@@ -72,48 +72,24 @@
 
             $feedItem['content_html'] = $item->getBody();
 
-            $feedItem['_indieweb'] = array();
+            $feedItem['_meta'] = array();
             if (method_exists($item, 'getMetadataForFeed')) {
-                $feedItem['_indieweb'] = $item->getMetadataForFeed();
-            } else if ($item instanceof \IdnoPlugins\Like\Like) {
-                $feedItem['external_url'] = $item->getBody();
-                unset($feedItem['content_text']);
-                if (!empty($item->repostof)) {
-                    $feedItem['_indieweb']['repost-of'] = $item->repostof;
-                    $feedItem['_indieweb']['type'] = 'repost';
-                } else if (!empty($item->likeof)) {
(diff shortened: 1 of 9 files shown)


templates/jsonfeed/shell.tpl.php
```diff
@@ -72,48 +72,24 @@
 
             $feedItem['content_html'] = $item->getBody();
 
-            $feedItem['_indieweb'] = array();
+            $feedItem['_meta'] = array();
             if (method_exists($item, 'getMetadataForFeed')) {
-                $feedItem['_indieweb'] = $item->getMetadataForFeed();
-            } else if ($item instanceof \IdnoPlugins\Like\Like) {
+                $feedItem['_meta'] = $item->getMetadataForFeed();
```
mapkyca: Any reason why these couldn't be pulled from jsonSerialize()? ... I'd prefer to avoid adding an extra method on classes if possible, or if it isn't possible to avoid, then we should probably add a sensible default to the Entity superclass, so we don't have to test for the method...

cleverdevil: Good question, @mapkyca. I looked into using `jsonSerialize()` for this purpose, but for the same reasons that I moved this serialization from `templates/json/shell.tpl.php` to `templates/jsonfeed/shell.tpl.php`, I think its best to keep these two things separate. While both methods are generating fields for serialization in JSON, they are doing so for very different purposes (API consumption vs. JSON Feed generation). The current uses of `jsonSerialize()` and the API may have different requirements from consumers of JSON Feed.

I can add a sensible default to `Entity` easily, since I already have the basis for it inside of  `templates/jsonfeed/shell.tpl.php`. I'll just move it. Stand by!

cleverdevil:
Hey @mapkyca and @benwerd, I've made some tweaks here, and am pretty happy with it at this point. Any feedback or input?

Issue #1772, opened 2017-06-01
Bonita template should be moved from /external


mapkyca:
## While trying to do this:

Since benwerd/bonita has been officially moved into Known, we should move it out of /external and update the readme accordingly...

## I encountered this error:

It's confusing, where do I patch?

## Some other notes:

/cc @benwerd

mapkyca:
Of course, where would we put it? Maybe move the directory "Bonita" to the top level, ditching all the other stuff like readme and examples...

PR #1773, opened 2017-06-01
Updating CSRF token generation to a more secure mechanism 

mapkyca:

Moved from sha1() hash to sha256 hmac

## Here's why I did it:

hmacs are cryptographically secure, so it's harder to fix the site secret (public key).

Holding off merge until status of #1772 has been decided.


mapkyca:
Actually, it looks like Bonita already contains some Known only hooks, so think it's safe to merge...

PR #1774, opened 2017-06-01
Token updating

mapkyca:

* When calling a page via XHR, a HEAD request will now return a new CSRF token.
* Added a helper JS function to use it.

## Here's why I did it:

To allow javascript only pages.

NOTE: Please check this through and read ticket discussion before merging, just incase I'm missing something and opening a horrible security hole...


mapkyca:
Closing, superseded by #1775

Issue #1790, opened 2017-06-23
m_value index with PostgreSQL causes long posts to misbehave


a3nm:
## While trying to do this:

Using PostgreSQL, try to publish a long post on Known (longer than a few kilobytes).

## I encountered this error:

When trying to access the post, the following error appears: "This content has been removed. This resource isn't here anymore."

The post contents can no longer be edited (although the post can be deleted).

## Some other notes:

In the Apache2 error log, the following is reported (where I replaced installation-specific parts by XXX):

> `[XXX] [:error] [pid XXX] [client XXX] SQLSTATE[54000]: Program limit exceeded: 7 ERROR:  index row size 2800 exceeds maximum 2712 for index "m_value"\nHINT:  Values larger than 1/3 of a buffer page cannot be indexed.\nConsider a function index of an MD5 hash of the value, or use full text indexing., referer: XXX`

The cause appears to be the `m_value` index created here: https://github.com/idno/Known/blob/master/schemas/postgres/postgres.sql#L89. This index will misbehave for posts that are too long because of a PostgreSQL limitation.

I was able to fix the problem by deleting manually the `m_value` index from PostgreSQL.

## Suggested fix

If this index is actually useful, it should be created with a different index type which does not have a size limitation, e.g., following https://stackoverflow.com/questions/41811468/postgresql-partial-index-exceeds-maximum-index-row-size. Alternatively, if this index is not useful, it should not be created.

mapkyca:
<a href="https://github.com/idno/Known/issues/1790" class="u-in-reply-to"></a>Hmmm... 

The value index is necessary for various metadata value searches, so probably needs to be kept.

I no longer have access to a postgres install, so it'd be really handy if you could submit a PR with a failing test that reproduces this issue..

a3nm:
Hi,

Thanks for your answer! Sorry but I'm not familiar at all with the Known
codebase, so it's not easy for me to write a test as you requested.
However, it should be fairly easy to reproduce this bug on a fresh Known
install with Postgres when trying to publish a long post.

Best,

-- 
Antoine Amarilli

mapkyca:
<a href="https://github.com/idno/Known/issues/1790" class="u-in-reply-to"></a>I tried to replicate using a pull request and the travis postgres build, but so far I've not been able to...

a3nm:
Thanks for trying it out! I have investigated too. TLDR: you may need a larger post size to reproduce the problem (try replacing 10240 by 2000000 in your PR).

[...]

PR #1803, opened 2017-06-30
Refs #1790: Attempting to replicate reported fault

mapkyca:

Added a test with a really long post

## Here's why I did it:

Attempting to replicate the fault reported in #1790


mapkyca:
This is a test, don't merge...

PR #1819, opened 2017-07-10
Renaming files to match community guidelines

mapkyca:

Renaming existing files to match github's community guidelines template.

## Here's why I did it:

Completeness, and to make sure github spotlights them to new contributors.


mapkyca:
Actually, I think this shouldn't be controversial...

PR #1936, opened 2017-10-30
Correctly check type when sending annotation via micropub

cweiske:
Also throw the error correctly.

- Check if the annotation type is correct
- Call the wrong-type-error function with the correct number of arguments
- Fetch post by slug not UUID, that does not work anymore.

## Here's why I did it:
Adding annotations via micropub (#1535) was broken.


mapkyca:
Thanks for the fix!

PR #1941, opened 2017-11-04
Upgrading TinyMCE

mapkyca:

* Upgraded TinyMCE
* Enable mobile theme
* Using native character count (since our counter didn't seem to work)

## Here's why I did it:

Newer version supports a mobile responsive theme for touch only devices


mapkyca:
For the tape, this PR was blocked by a bug in the upstream project, which afaik hasn't been fixed yet.

PR #1964, opened 2017-11-27
Allow stub console plugins

mapkyca:

Allow php files in ConsolePlugins/ to be scanned for console plugins

## Here's why I did it:

It was time consuming to create full directory hierarchies when you just needed a quick console file.

### DIFF
--- a/known.php
+++ b/known.php
@@ -72,6 +72,18 @@
                                 ->setCode([$c, 'execute']);
                     } 
                 }
+            } else {
+                $stubclass = "ConsolePlugins\\$file";
+                $stubclass = str_replace('.php', '', $stubclass);
+                if (class_exists($stubclass)) {
+                    $c = new $stubclass(); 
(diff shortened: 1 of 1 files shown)


benwerd:
I wonder if this should be true for `IdnoPlugins` too?

mapkyca:
Maybe, although "proper" plugins are always at least two files - the .php and the manifest, so we'd need to also provide a way of accounting for that.

benwerd:
Oh, you're right. Never mind. I was thinking about consistency from a developer standpoint, but I don't think removing the manifests would be a great idea, fwiw.

mapkyca:
Aye, and in any case I think we should "encourage" compartmentalisation in the main plugin hierarchy. It makes sense to allow stubs for console since they're almost always a single file, and I don't anticipate a massive ecosystem for console plugins..

benwerd:
Yeah, I think most people don't know they exist, and might not be able to use them - which is fine.

PR #1965, opened 2017-11-28
Somehow some deploys have a composer lock file

mapkyca:

Added composer.lock to the blocked list

## Here's why I did it:

Some deploys seem to use composer for some stuff, block the lock file from showing.


benwerd:
Possibly all .lock files?

mapkyca:
.. might be smart...

PR #1977, opened 2018-01-19
Resolving an issue with the UrlUnfurler in PHP 7.2

cleverdevil:
PHP 7.2 now reserves the word "Object." Rename to "BaseObject" to fix.

Renamed "Object" to "BaseObject."

## Here's why I did it:

URL Unfurling was failing on PHP 7.2 with an error:

```PHP Fatal error:  Cannot use 'Object' as class name as it is reserved in /var/www/cleverdevil.io/Idno/Entities/Object.php on line 12```

### DIFF
--- a/Idno/Common/ContentType.php
+++ b/Idno/Common/ContentType.php
@@ -11,7 +11,7 @@
 
             // Property containing the entity class associated with this content type (default is generic object type)
             static public $registered = array();
-            public $entity_class = 'Idno\\Entities\\Object';
+            public $entity_class = 'Idno\\Entities\\BaseObject';
             public $handler_class = 'Idno\\Common\\ContentType';
             public $title = 'Content type';
             public $indieWebContentType = array();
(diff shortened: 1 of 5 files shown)


mapkyca:
I'm surprised this hasn't been handled with namespacing, since Object isn't the full class name... 

Renaming the class is the obvious solution, but I'd be interested to see if there's another possibility since a lot of things inherit off Object...

mapkyca:
Another thing that's curious is that it's only spitting out this error here, given that pretty much everything inherits off Object. 

I wonder if this might be a false +ve while 7.2 is still in development, might be worth reporting...

mapkyca:
Doing a bit of background reading at it seems at it seems there's a whole bunch of projects running into this one. Looks like this is going to be a "thing" now...

mapkyca:
... going to try and get a PHP 7.2 environment set up in the next week or so and try and get into this...

mapkyca:
Sticking a pin in this for now, I think it's a larger issue and needs to be managed carefully...

PR #2067, opened 2018-04-09
Adding diagnostics for detecting the latest version off of github.

mapkyca:

Adding diagnostics for detecting the latest version off of github.

## Here's why I did it:

Attempting a more responsive mechanism for detecting new releases of Known, this time keying off GitHub.

This is designed to be temporary, and not to replace the official release/update notification.


mapkyca:
Discussion, do we actually want to do this with the main check?

PR #2069, opened 2018-04-13
Remove token debug

mapkyca:

Removing the token debug logging

## Here's why I did it:

The situation that needed this has been resolved, and logs that are too verbose are hard to read.


srcr:
Would it be an idea to do the same for the file Idno/Pages/Service/Security/CSRFToken.php line 16, this one is still generating CSRF token debug info.

mapkyca:
Ahh... good spot, many thanks!

PR #2095, opened 2018-06-08
Update icons post font-awesome upgrade.

cleverdevil:

Changed the icon for "Like" from a missing icon to an appropriate new icon from font-awesome, and added an icon for "Bookmark" content to make things more consistent.

## Here's why I did it:

@mapkyca did an awesome thing and upgraded font-awesome. This is one of the only side-effects in core.


mapkyca:
Nice spot! Thanks for helping clear up the sawdust :)

PR #2143, opened 2018-09-09
Composer package

mapkyca:

Added an initial composer.json

## Here's why I did it:

In line with the roadmap, this is so Known can eventually be added to packagist, and installable via composer.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -30,6 +30,10 @@
   Order allow,deny
   Deny from all
 </Files>
+<Files ~ "composer\.json">
+  Order allow,deny
+  Deny from all
+</Files>
 <Files ~ "package\.json">
   Order allow,deny
   Deny from all
--- a/composer.json
+++ b/composer.json
@@ -6,5 +6,10 @@
 	"php": ">=7.0"
     },
     "license": "Apache-2.0",
-    "keywords" : ["blogging", "publishing", "social"]
+    "keywords" : ["blogging", "publishing", "social"],
+    "support": {
+	"forum": "https://groups.google.com/forum/#!forum/known-dev",
+	"docs": "http://docs.withknown.com",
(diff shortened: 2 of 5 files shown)


physcocode:
just a suggestion why not we start maintaining our external dependencies from composer from this pull request itself

mapkyca:
Eventually, yes. But right now that's a fair amount of work I don't have time for. Additionally there's the goal of not *requiring* composer... I do think that we should still have a working package out of the box.

Plus I know a couple of packages there have or at least had some custom mods, so we'll need to audit those.

physcocode:
1) we can package known with all composer library included so that composer is not required mandatory 
2)It would help in maintaing and upgrading versions , nor more outdated libraries 
3) we should NEVER modify external libraries as it makes harder to upgrade when new version arrives , we should rather do it via interfaces

mapkyca:
1. Mayhaps, but I'd prefer not to *require* apt-get rootkit for people just checking out the git repo ;)

2. I prefer pegged versions, since blind updates cause problems, but in principle I agree.

3. Agreed, but these cases were when the library was either not updated and had to be adopted, or when the upstream project didn't provide a minified version of whatever. There was also a forked version of the mention client that was necessary for a time, but that has since been tidied. Basically, it's probably fine, since much of this has been tidied up over time, but we need to run through and double check.

physcocode:
Maybe you interpreted 1 point wrongly , I meant we can include the "vendor" folder inside our release so that anyone downloading the release package dosnt have to care about composer

mapkyca:
That's fair, and basically what we've got right now with /external. 

I do think moving dependencies over is probably wise, but I'd like to do this as a staged process in a later PR. For one thing there are a couple of projects which I don't believe are currently available via composer, so while we sort that our external and vendor will probably have to co-exist for a time.

PR #2146, opened 2018-09-11
Improved installer to better handle .htaccess

mapkyca:

* Merge rather than concat webserver configs
* Add a backup mechanism for existing configs so it's easier to fix screwups
* Housekeeping and removal of duplicate / unused functionality

## Here's why I did it:

The installer didn't handle cases where the .htaccess was already written very well.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,4 +1,4 @@
-## BEGIN Known .htaccess file
+## BEGIN Known Webserver Config
 
 <IfModule mod_rewrite.c>
     RewriteEngine on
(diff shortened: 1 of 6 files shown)


.htaccess
```diff
@@ -90,3 +94,5 @@
 AddType video/ogg .ogv
 AddType video/mp4 .mp4
 AddType video/webm .webm
+
+## END Known Webserver Config (don't remove)
```
benwerd: Let's add another newline here.

mapkyca: Probably best duplicate this in htaccess.dist and htaccess-2.4.dist

PR #2150, opened 2018-09-12
Adding gruntfile to plugins

mapkyca:

Adding a gruntfile to core plugins

## Here's why I did it:

To lint and minify built in javascript, and to make it easier to generate language files


benwerd:
LG! Also, someone just reached out to me asking to translate the whole thing into French yesterday. I wonder if we should have a guide for that in the docs, too.

mapkyca:
There is a guide already! http://docs.withknown.com/en/latest/developers/languages/

PR #2157, opened 2018-09-15
Redesigning async message queues

mapkyca:

Instead of the service pulling objects from the database directly, I have added a api service endpoint for the queue's functionality and the service now calls those.

## Here's why I did it:

Maintaining an open connection to the database was problematic, especially for Mongo. This approach should be more reliable. In addition it provides scope for having the queue management service called by something else - e.g. javascript, etc.


mapkyca:
Actually, going to hold off on this... I think I got ahead of myself.

In my excitement, I forgot the main reason why this was run as a CLI service - execution time and memory. We want to be able to palm off things that will take a long time and/or a lot of resources (e.g. transcoding videos) to a background task. Having these run as part of the web server process (via an API call) is a bad plan.

PR #2159, opened 2018-09-16
Redesigning async message queues (second attempt)

### LINKED ISSUE #2155: Rethink / rewrite async queue service
## While trying to do this:

Running existing asynchronous message queue

## I encountered this error:

Service (at least on mongo) will periodically terminate, typically with:

``` No suitable servers found (`serverSelectionTryOnce` set): [Server closed connection. calling ismaster on 'localhost:27017']Unknown command error```

This appears to be uncatchable, and so re-establishing a connection appears to not be possible. 

We should probably integrate with an existing message queue service (RabbitMQ or something), or have a service that could ping an execution endpoint (that will unpack the queued task and execute), rather than constantly running a Known instance.

### LINKED ISSUE #2156: Rethink / rewrite Cron service
## While trying to do this:

Run the cron service

## I encountered this error:

Service will, on mongo at least, terminate periodically.

## Some other notes:

Refs #2155

### PR DESCRIPTION
**Pull request by mapkyca**

This is a second attempt at redesigning the async message queues. It contains many of the changes from #2157, but has a different approach to dispatching.

In addition to the changes in #2157, I have introduced a queue management CLI command to list and dispatch events, and removed the dispatch API endpoint for the time being (because dispatching expensive events in an api call represents a DoS vector.

The queue service now makes use of the management API to retrieve a list of events pending, as well as to garbage collect, however actual dispatch is made using a system call to the event management tool.

## Here's why I did it:

The queue management API calls introduced in the previous attempt are actually quite useful, since for one thing it removes the requirement to maintain an open DB connection. 

Dispatching the event using the event management tool now means that we can run the dispatch as a CLI process (keeping the advantages of that), but since it now runs as a separate process, any database connections are single use.

Keeping the management api, and introducing a management tool also means that in future we can rewrite the queue service to be driven by something else - node, whatever.

Closes #2155 
Closes #2156

### CODE REVIEW

**Comment by benwerd** [comment]
This and its companion are pretty big PRs - I didn't get a chance to review over the weekend but will tonight.

**Comment by mapkyca** [comment]
Second PR also includes this one, so may be the easiest one to review.

File: ConsolePlugins/EventQueueService/Main.php [hunk 1]
```diff
@@ -67,17 +45,18 @@ public function execute(\Symfony\Component\Console\Input\InputInterface $input,
...
-                                    foreach ($events as $evnt) {
+                                
+                                if ($events = \Idno\Core\Service::call('/service/queue/list/')) {
+                                    foreach ($events->queue as $event) {
                                         try {
-                                            $eventqueue->dispatch($evnt);
+                                            \Idno\Core\Idno::site()->logging()->info("Dispatching event $event");
+                                            //\Idno\Core\Service::call('/service/queue/dispatch/' . $event);
+                                            
+                                            system(escapeshellcmd("./known.php event-queue-manage $queue dispatch $event"));
```
benwerd: Not every environment can do this. It's possible to disable this ability, and many hosts do - so we need to at least check for that and fail in a way that doesn't bring Known down with a technical message. (i.e., it's got to be non-catastrophic for those users.) I don't know if that's the case here, but if it is, we need to be able to return a nice "your hosting environment doesn't support this action" sort of message.

Idno/Core/AsynchronousQueue.php
```diff
@@ -8,6 +8,12 @@
  */
 class AsynchronousQueue extends EventQueue
 {
+    function registerPages() {
+        \Idno\Core\Idno::site()->addPageHandler('/service/queue/list/?', '\Idno\Pages\Service\Queues\Queue');
+        //\Idno\Core\Idno::site()->addPageHandler('/service/queue/dispatch/([A-Za-z0-9]+)/?', '\Idno\Pages\Service\Queues\Dispatch');
```
benwerd: Let's just kill this line.

Idno/Pages/Service/Queues/GC.php
```diff
@@ -0,0 +1,28 @@
...
+        
+        public function getContent() {
+            
+            Idno::site()->template()->setTemplateType('json');
+            
+            \Idno\Core\Service::gatekeeper();
+            
+            $eventqueue = \Idno\Core\Idno::site()->queue();
+            if (!$eventqueue instanceof \Idno\Core\AsynchronousQueue) 
+                throw new \RuntimeException("Service can't run unless Known's queue is Asynchronous!");
```
benwerd: What does this error message mean? Is there a clearer way we can say this?

benwerd:
@mapkyca did you see. my comments on the review?

ConsolePlugins/EventQueueService/Main.php
mapkyca: It only affects console services, specifically *this* console service. For now, I think the people who want or need asynchronous queues will be few. 

I'll look to improve this, but I do think a system call is going to be necessary - running in the same process is flakey.

mapkyca: A nice warning is probably good enough for now though...

Idno/Pages/Service/Queues/GC.php
mapkyca: It actually probably isn't needed at all in this context. Copy/paste error :)

mapkyca:
Yes, I'm applying the feedback as a separate PR to make management of the merges easier

PR #2161, opened 2018-09-18
i18n: Call registration hook for translations in console plugins

mapkyca:

Adding a registerTranslations() hook to console plugins

## Here's why I did it:

Looking to the future, because console commands should probably be translatable too.


mapkyca:
ok...

but why?

PR #2169, opened 2018-09-20
Revised mkdocs configuration

benwerd:

It turns out we didn't need the pages configuration option in the mkdocs.yml file. I removed it, and have tested that mkdocs parses and renders the documentation OK.

## Here's why I did it:

Docs haven't been regenerated publicly for a long time.


benwerd:
It did, once upon a time. I've tested this locally, yes.

PR #2171, opened 2018-09-20
Shrinking controls for unfurl to icons only

benwerd:
The text labels were huge, and unfortunately inconsistently sized across templates. Here's how the revised version looks:

screen shot 2018-09-19 at 10 17 01 pm

### DIFF
--- a/templates/default/content/unfurl.tpl.php
+++ b/templates/default/content/unfurl.tpl.php
@@ -8,8 +8,8 @@ if (empty($vars['object']->hide_preview)) { ?>
     <div class="unfurl col-md-12" style="display:none;" data-url="<?= $dataurl; ?>"></div>
     <?php if ($vars['object']->canEdit() && (!empty($vars['object']->_id))) { ?>
     <div class="unfurl-edit pull-right small">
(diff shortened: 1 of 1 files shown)


templates/default/content/unfurl.tpl.php
```diff
@@ -8,8 +8,8 @@
     <div class="unfurl col-md-12" style="display:none;" data-url="<?= $dataurl; ?>"></div>
     <?php if ($vars['object']->canEdit() && (!empty($vars['object']->_id))) { ?>
     <div class="unfurl-edit pull-right small">
-        <a href="#" class="refresh"><i class="fa fa-sync"></i> <?= \Idno\Core\Idno::site()->language()->_('Refresh preview'); ?></a> &nbsp;
-        <a href="#" class="delete"><i class="fa fa-trash"></i> <?= \Idno\Core\Idno::site()->language()->_('Remove preview'); ?></a>
+        <a href="#" class="refresh"><i class="fa fa-sync"></i></a>
```
mapkyca: What about having the labels appear as hover titles?

My concern would be that people might confuse rubbish with delete post.

Other than that, LG

PR #2172, opened 2018-09-20
Adding hover descriptive title

mapkyca:

Added descriptive titles to url preview delete and refresh button

## Here's why I did it:

Refs #2171

### DIFF
--- a/templates/default/content/unfurl.tpl.php
+++ b/templates/default/content/unfurl.tpl.php
@@ -8,8 +8,8 @@ if (empty($vars['object']->hide_preview)) { ?>
     <div class="unfurl col-md-12" style="display:none;" data-url="<?= $dataurl; ?>"></div>
     <?php if ($vars['object']->canEdit() && (!empty($vars['object']->_id))) { ?>
     <div class="unfurl-edit pull-right small">
-        <a href="#" class="refresh"><i class="fa fa-sync"></i></a>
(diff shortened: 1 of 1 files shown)


templates/default/content/unfurl.tpl.php
```diff
@@ -8,8 +8,8 @@
     <div class="unfurl col-md-12" style="display:none;" data-url="<?= $dataurl; ?>"></div>
     <?php if ($vars['object']->canEdit() && (!empty($vars['object']->_id))) { ?>
     <div class="unfurl-edit pull-right small">
-        <a href="#" class="refresh"><i class="fa fa-sync"></i></a>
-        <a href="#" class="delete"><i class="fa fa-trash"></i></a>
+        <a href="#" class="refresh" title="<?= \Idno\Core\Idno::site()->language()->_('Refresh preview'); ?>"><i class="fa fa-sync"></i></a>
+        <a href="#" class="delete" title="<?= \Idno\Core\Idno::site()->language()->_('Remove preview'); ?>"><i class="fa fa-trash"></i></a>
```
benwerd: Do the language methods automatically escape strings? If not, we should probably escape this. And think about an automatic way to do that.

mapkyca: No, they don't automatically escape strings. There are times where this is not desirable, so I don't think this should be applied across the board.

For the record, WordPress handles this by having several i18n string methods so you can decide which to use.

Issue #2177, opened 2018-09-21
Under some circumstances Service signatures don't match up


mapkyca:
## While trying to do this:

Run the service queue on mapkyca.com

## I encountered this error:

Known (mapkyca.com): error - Sorry, signature doesn't match up.

## Some other notes:

Give us some context: 

* It'd also be really handy if you could tell us the contents of your ```version.known``` file

2018092001

* What database are you using? (e.g. mongo, mysql, postgres)

Mongo

PR #2178, opened 2018-09-21
Adding Service signature debug

mapkyca:

Adding debug to Service signature generation and gatekeeper

## Here's why I did it:

Under certain circumstances it seems that signatures fail to generate correctly.

Refs #2177


mapkyca:
I'm wondering whether this is a recurrence of the token generation we had before with multiple config elements.

mapkyca:
No longer required, replaced by #2180

PR #2195, opened 2018-09-24
Indieweb ox

mapkyca:

Some various bits and pieces noticed while at IWC Ox


mapkyca:
Direct merge proved to be a PitA, so have put together a second patch. See #2202

Issue #2221, opened 2018-10-04
Idno\Core\Idno::languages undefined in Checkin plugin


cleverdevil:
## While trying to do this:

Loading a Checkin post.

## I encountered this error:

Exception: 

```[04-Oct-2018 23:50:17 UTC] Known (cleverdevil.io): critical - Call to undefined method Idno\Core\Idno::languages() [/var/www/cleverdevil.io/IdnoPlugins/Checkin/Checkin.php:10]```

That line makes a call:

```return \Idno\Core\Idno::site()->languages('Checked into %s', [$this->placename]);```

That doesn't seem to work.

## Some other notes:

I'm on latest from master.

```version = "0.9.9-a"
build = 2018100101```

cleverdevil:
This seems to be fixed by changing "languages" to "language". Typo?

PR #2222, opened 2018-10-04
Fix typo

cleverdevil:

Fixing what I believe may be a typo.

## Here's why I did it:

See #2221.


benwerd:
Good catch! Thank you!

PR #2225, opened 2018-10-07
Adding JSON+LD support

mapkyca:

Adding JSON+LD support

Adds json+ld for site itself (including search function in google), user (including linking to social media posts), blog post, photos and media.

## Here's why I did it:

Structured data support is sort of the thing these days.

I'd love @edent 's input on this, as I understand this is right up your ally.

### DIFF
--- a/Idno/Common/Entity.php
+++ b/Idno/Common/Entity.php
@@ -503,12 +503,13 @@ namespace Idno\Common {
 
             return get_class($this) . ' ' . $this->_id;
         }
-        
+
         /**
          * Return the created timestamp of the entity.
          * @return unix timestamp
          */
-        function getCreatedTime() {
+        function getCreatedTime()
+        {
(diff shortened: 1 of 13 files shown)


edent:
Generally looks good. A couple of points.

` "@context" => "http://schema.org",` I think that should be https. Probably doesn't make a huge difference, but foot to be consistent.

Some of the data should include dates.

I'm a fan of microdata, but doesn't make too much difference. 

You can test using:
https://search.google.com/structured-data/testing-tool

mapkyca:
Thanks for your help!

mapkyca:
I've run it through the google tool, but yeah testing is good.

mapkyca:
This might be helpful: https://www.npmjs.com/package/wct-structured-data-testing (grunt task, not phpunit tho)

PR #2239, opened 2018-10-21
Remove Journal plugin

egoexpress:

Removed Journal plugin from the plugin list.

## Here's why I did it:

Tino Kremer removed the repo for the Journal plugin from GitHub.


benwerd:
Thank you!!

PR #2241, opened 2018-10-22
Improve form documentation code

mapkyca:

* changed form-fields to formFields
* Better handling of arrays

## Here's why I did it:

So that the api can be more easily parsed.


mapkyca:
@benwerd forgot to ask for review on this one!

PR #2247, opened 2018-10-25
Attempt to mimic PHP's array name conversion for JSON blobs

mapkyca:
Attempt to mimic PHP's array name conversion for JSON blobs
## Here's why I did it:

PHP magically converts ```foo[]``` into ```$_REQUEST['foo'] => array(...)```, and these names are expected by the code (e.g. for syndication).

This code attempts to mimic this functionality by automatically removing the "[]" from variable names. Actual [] blocks in payload should be escaped and not modified, but report any problems.

### DIFF
--- a/Idno/Common/Page.php
+++ b/Idno/Common/Page.php
@@ -198,6 +198,7 @@ namespace Idno\Common {
             // First, let's see if we've been sent anything in form input
             if (!empty($_REQUEST['json'])) {
                 $json = trim($_REQUEST['json']);
+                $json = str_replace('[]"', '"', $json); // Fake PHP's array conversion
                 if ($parsed = @json_decode($json, true)) {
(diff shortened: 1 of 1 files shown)


benwerd:
Huh! Can you give an example?

mapkyca:
Yes, testing to come. Out of time / really tired, now.

As for an example:

Syndication is POSTed as a variable "syndication[]"....  When this is traditional urlencoded/form data PHP magically converts it into "syndication" => [ foo, bar], which is what getInput and everything using the variable expects.

JSON blobs don't do this, so "syndication[]" gets converted to 'syndication[]' => [foo, bar]. Meaning anything looking for the variable can not find it.

benwerd:
Aha. Creepy and weird. Thanks for fixing.

PR #2248, opened 2018-10-25
Further composer changes

benwerd:


benwerd:
Oh FFS. Only consider the last commit please ...

Issue #2209, opened 2018-09-29
Enable IndiePub plugin by default (IndieAuth + Micropub)


aaronpk:
## While trying to do this:

Set up a new Known site and use a Micropub app

## I encountered this error:

App did not find a way to authenticate me

## Some other notes:

Enabling the IndiePub plugin then works.

jaduncan:
I would also recommend this.

cleverdevil:
Totally agree, I'd like to see Micropub and IndieAuth be enabled by default, and promoted as the "native" way to publish content and interactions via third-party apps like Quill, Together, Indigenous, and Omnibear.

jgmac1106:
Yes I tried to set my kids up with indigenous so they could post and check out each other' blog.

+1 as well for 1.0

mapkyca:
<a href="https://github.com/idno/known/issues/2209" class="u-in-reply-to"></a>See discussion on #2249. Indiepub plugin is enabled by default, but it's likely an interaction with older versions or saved configs.

Issue #1846, opened 2017-08-05
Admin (and maybe settings) should probably have it's own page shell


mapkyca:
## While trying to do this:

Use admin settings

## I encountered this error:

I had broken something on the theme which meant all the forms were broken. Currently it's all too easy for a theme or plugin to break something and leave you with an inoperable admin section. 

Discuss

## Some other notes:

Ref #1641

mapkyca:
Based on discussion, I'm bumping this and adding it to the Known 1.0 todo list.

This will likely be a breaking change for a number of plugins, so should probably be done now rather than later, and I think it is important to follow the wordpress model and keep settings and admin settings separate from any code added by the theme (which can break the site)

physcocode:
Is it about themes should not override admin view ? Like WordPress?

mapkyca:
<a href="https://github.com/idno/known/issues/1846#issuecomment-444234782" class="u-in-reply-to"></a>Basically. I'd like to see admin and user settings in its own isolated theme. We've had issues where people have for example embedded javascript into their theme (or fudged their PHP) and it renders the whole site unusable and making it difficult to disable the theme and work backwards.

PR #2249, opened 2018-10-29
Adjusting initial plugins

benwerd:
Removing Firefox and taking the style / footer JS plugins out of the default stack.

### DIFF
--- a/Idno/Core/Config.php
+++ b/Idno/Core/Config.php
@@ -19,7 +19,7 @@ namespace Idno\Core {
             'sessionname'            => 'known', // Default session name
             'boolean_search'         => true, // Should search be boolean?
             'open_registration'      => true, // Can anyone register for this system?
-            'initial_plugins'        => array('Status', 'Text', 'Photo', 'Event', 'Firefox', 'FooterJS', 'IndiePub', 'Styles'),
+            'initial_plugins'        => array('Status', 'Text', 'Photo', 'IndiePub'),
             'plugins'                => array(), // Default plugins
             'assets'                 => [      // Assets to be included
                                                'mediaelementplayer' => true,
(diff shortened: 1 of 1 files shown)


mapkyca:
You also removed event, was this deliberate?

Also also, might be worth looking at #2209 - which looks like to be the case, but might be worth investigating why they're reporting it's not active by default.

benwerd:
It was deliberate. Event is kind of shoddy. I'm inclined to pop it out into its own repo.

And yep. Weird that it wouldn't be on by default. Strongly suspect it's a difference between the public package and current master.

mapkyca:
Rather see PRs than it getting bumped tbh, things that end in their own repos tend to end up falling off people's radar... certainly my own!

benwerd:
For sure. I think I'm considering just wiping it; my take is, the fewer features to maintain, the better. Better to have a few core things that work really really well than a million that are kind of dubious.

jgmac1106:
+1 to smaller core and allowing community plugins though I must say Known has gotten closest to a correct event better than any platform. Still an event indo plugin would be fine

mapkyca:
Maybe put it up for adoption then. I don't like the idea of deleting code, and I know some people still use some of the functionality - e.g. rsvps

mapkyca:
Also yeah, I think event isn't terrible. I've certainly done a fair amount of work re timezones. Probably just needs a bit of spit and polish

jgmac1106:
In terms of being used yes most HWCs use somebody's Known site for events
and to accept RSVPs as Known is only event UI that works.

mapkyca:
My vote would be to keep events in core, and instead knock off the rough edges. There are other plugins I'd get rid of or streamline first - e.g. comic. 

I'd also be tempted to spin off footerjs etc (which has the power to break the site), with a view to replacing this with something better, as well as build #1846

jeremycherfas:
+1 for leaving Events in, even if it does have rough edges. It also works, which is good even if the code isn't super smooth.

PR #2250, opened 2018-10-31
Removing Comic plugin

mapkyca:

Removing Comic plugin

## Here's why I did it:

In an attempt to make Known more streamlined as we move towards 1.0, removing Comic plugin (which is currently pretty much a poorly maintained and featureless version of the Photo plugin) from core and placing it in it's own repo.

To obtain the comic plugin from now on please look at https://github.com/idno/Comic


physcocode:
+1

PR #2272, opened 2018-11-18
Adding a CLI export method

mapkyca:
Adds a CLI export function to match the import function, for similar reasons, that CLI has fewer restriction in terms of runtime and memory.

Improved the export function somewhat in order to make it export more safely, and handle errors better.

Still no import for this, PRs welcome.

Discussion: My thinking is that we should actually remove the web interface import and export code, since this is a pretty technical thing to do. Raising the bar slightly to using the CLI (which works better anyway as you're not going to hit so many memory issues) might reduce some of the help questions. Anyway.

### DIFF
--- a/ConsolePlugins/Export/.gitignore
+++ b/ConsolePlugins/Export/.gitignore
@@ -1,0 +1,3 @@
+node_modules
+package-lock.json
+!package.json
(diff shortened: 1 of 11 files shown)


benwerd:
I don't agree about removing the interface, btw - but it definitely needs to be rewritten. It's a technical thing to do but we've got to lower that barrier somehow.

PR #2277, opened 2018-11-20
Update Endpoint.php

jgmac1106:
fixing syndication target for micropub clients per: https://github.com/swentel/indigenous-android/issues/152#issuecomment-438382313

PR #2278, opened 2018-11-20
Changing syndication targets to arrays

voss:
In reference to https://github.com/swentel/indigenous-android/issues/152

Indigenous was throwing errors when using known as it expects syndication targets to be an array, and not string.

I changed this on my own instance of known 0.9.9-a (build = 2018101001) and Indigenous can now read the syndication targets correctly.

I also checked using Quill, and Aperture, where the syndication targets also work.


benwerd:
This seems to be a duplicate of #2277. Closing here and approving there! But thank you!

voss:
> This seems to be a duplicate of #2277. Closing here and approving there! But thank you!

Yes, I and @jgmac1106 should have coordinated better (:

PR #2295, opened 2018-12-12
Photo lightbox "fixes"

tincho:

(Hi there) I made some changes directly to Photo.tpl.php 

- `data-gallery` attr includes the post Id so it's more unique
- if the plugin is enabled:
  - adds `data-toggle="lightbox"` attr so ekko-lightbox can be queried unequivocally, and more efficiently
  - the link's `src` attr points straight to the image file

## Here's why I did it:

Originally I had found a bug: when a post featured multiple photos, Lightbox would show up for each, but not include all of them in the same modal *as a gallery*.... having to close the modal for each and reopen clicking the next. 
Actually there's some JS code that replaces the `src` attr right before triggering the Lightbox, but it's not working (at least for me).
In the theme I'm working on, overrided `footer.tpl.php` so the Lightbox is setup like this:
```javascript
$(document).on('click', '[data-toggle="lightbox"]', function(event) {
	event.preventDefault();
	var $link = $(this);
	$link.ekkoLightbox({ type: 'image' });
});
```

I know Lightbox is shipped separately but it's almost a "core" (must have IMO!) plugin. Making these modifications there would implied having to override .

Probably will soon provide working "demo"


tincho:
Hi there! 
It's been a while, but I guess there's no need to check for Lightbox nor setting `data-toggle=lightbox` attr. 
Setting always `data-remote` should be enough for Lightbox to detect this. 
I'll try again in a clean install and update this PR

mapkyca:
Be sure to check out @benwerd's recent changes in #2320, which I think were put in to help you with this sort of stuff...

mapkyca:
Any updates on this, and the feedback?

mapkyca:
Closing for now, please reopen once you've looked at the feedback!

tincho:
Hi there! That's OK.
Excuse me the lack of follow up. I've been working on other stuff lately and still haven't deployed my Known-based publishing site. Will catch up soon probably.

PR #2301, opened 2018-12-17
Comments plugin: IDs were not unique causing errors

tincho:

Replaced id="comment-form" with id="comment-form-(last 5 digits of uuid)" in `comments/public/form.tpl.php` and fix the related JS.

## Here's why I did it:

Specifically I detected this in the /profile/username page, this was causing bugs because IDs were being repeated from post to post. Maybe somewhere else the same error could occur.


benwerd:
Thank you!

Issue #2141, opened 2018-09-06
Split Media into Audio and Video plugins


benwerd:
Video is, of course, much much much harder to implement.

But these two media types have very different needs and should be two different plugins.

mapkyca:
Yes, probably smart.

For video it might be worth merging in the transcoding work I did elsewhere (https://github.com/mapkyca/KnownVideoTranscode)

benwerd:
I'd be very against merging use of ffmpeg (etc) into core; the idea of supporting that on most systems gives me the shivers. I suspect this is a case where a third-party hosted service or an external plugin could do the gruntwork, and then post into the Known stream via integration or webhook.

mapkyca:
<a href="https://github.com/idno/Known/issues/2141#issuecomment-419240112" class="u-in-reply-to"></a>Fair point. My understanding of this ticket was that we were talking about spinning out Video into its separate third party plugin (rather than having it as a legacy feature / easter egg in Audio). Since video is still something that requires transcoding (although browser support *is* getting better) it makes sense to bundle ffmpeg into that video plugin and make that "complete", but to not necessarily support it ourselves.

PR #2336, opened 2019-02-07
Rebranding the legacy "audio" plugin as a "media" plugin. 

cleverdevil:
This effort was halfway done, and this change should bring it all the way there.

Renamed "audio" in relevant places to "video" and picked a more appropriate icon.

## Here's why I did it:

Inconsistency makes me sad.

## Checklist:

- [ ] I've added tests where applicable

### DIFF
--- a/IdnoPlugins/Media/ContentType.php
+++ b/IdnoPlugins/Media/ContentType.php
@@ -5,8 +5,8 @@ namespace IdnoPlugins\Media {
     class ContentType extends \Idno\Common\ContentType
     {
 
-        public $title = 'Audio';
-        public $category_title = 'Audio';
+        public $title = 'Media';
+        public $category_title = 'Media';
         public $entity_class = 'IdnoPlugins\\Media\\Media';
         public $indieWebContentType = array('media','audio','video');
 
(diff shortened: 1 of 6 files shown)


mapkyca:
Hmm... IIRC Audio used to be called Media, but it was changed since uploading video isn't reliable. Even to this day you still need to do a lot of transcoding to make something that's web compatible across browsers. 

I think the long term plan is to split this "media" plugin into audio and video, package audio and then keep video as an optional download for techy folk.

Right now, my vote would be to keep it as is, with video support as a hidden easter egg for people who are willing to take on the overhead of making sure their video is web ready, but I'll let @benwerd have the final say...

cleverdevil:
I hear you, but right now it’s sort of in a halfway state, and is confusing if you upload a video, because it’s still listed under a category of “audio.” I’d rather just be up front that it supports both, but that the video bit has caveats or is in beta or testing or whatever. Also happy to help make it work better if we have specific cases where it’s broken or buggy.

mapkyca:
Hmm... given a choice, I'd actually opt for removing references from video entirely. Like I said, having video work reliably isn't all that easy, so to keep things simple for non-technical users I'd prefer to move any video handling into it's own community supported plugin.

See discussion on #2141

cleverdevil:
Good points @mapkyca and thanks for sharing the discussion. I'll await feedback from @benwerd, but would be happy to either work on getting this PR polished for merging into core, or to be given the directive that a community plugin would be a better path, and then to work with @mapkyca on creating a new plugin that is solely for video and includes his awesome transcoding features.

mapkyca:
Bump at @benwerd

mapkyca:
Closing for now

PR #2353, opened 2019-02-24
Composer install (PARTIAL)

mapkyca:

Moving what I can from /external over to composer

## Here's why I did it:

Because it's $CurrentYear, and there are advantages.

This is a PARTIAL conversion, converting most stuff to composer. Remaining stuff is either customised by us, has no clear version to migrate to, or (like mongo) needs some shims on our side to work correctly. So, a migration plan needs to be worked out for these.

Refs #1064

## Checklist:

- [ ] I've added tests where applicable


mapkyca:
All changes that are easy to do have been done - other codes don't have clear versions to use.

mapkyca:
Hmm... I updated the installation docs. Most people *seem* to install from packages, and for tech users I've got a fatal error that prompts for composer install.

Probably good to have in the readme as well...

mapkyca:
Updated

benwerd:
I think we’ll need to say more - this is literally the bare minimum we could say - but this will work for now.

mapkyca:
Weirdly git seems to have screwed up, and I made the changes but cant push them.

mapkyca:
Come back to the docs - though non-tech users are going to use premade packages

PR #2400, opened 2019-04-18
Consistently use getStaticURL for js/css/img assets

lindner:

Replaced usages of `config()->getDisplayURL()` and `config()->url` that serve images, javascript, or css with `config()->getStaticURL()`

## Here's why I did it:

- Address consistency issue
- Allow for performance wins that come from CDN usage.

## Checklist:

- [ ] I can run the unit tests successfully.

### DIFF
--- a/Idno/Common/Entity.php
+++ b/Idno/Common/Entity.php
@@ -1007,7 +1007,7 @@ namespace Idno\Common {
                 return $page->getIcon();
             }
 
-            return \Idno\Core\Idno::site()->config()->getDisplayURL() . 'gfx/logos/logo_k.png';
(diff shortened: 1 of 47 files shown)


lindner:
Looks like the failure is for the php7 build that can't find phpunit

Might be wise to move phpunit to composer so you get consistent results from tests.

lindner:
Updated with gpg signed commit.

mapkyca:
Urgh... void type hinting isn't available in php 7, so travis tests are failing.

Going to remove 7.0 from the build for now, since I don't want to maintain two versions of the tests and this will sort itself out before too long. 

Thanks for the patch, looks good.

lindner:
Thanks, looking forward to an approval.  Looks like I can't restart the Travis process; but you can see that the failure is unrelated to the change (only failed on PHP 7).

Issue #2409, opened 2019-04-24
AllowOverride None detection no longer works


esavir:
## While trying to do this:
I was trying to install Known.

## I encountered this error:
"Rewriting appears to be disabled. Usually this means "AllowOverride None" is set in apache2.conf which prevents Known's .htaccess from doing its thing. We tried to fetch a URL that should redirect to default.js"
"You can usually fix this by setting AllowOverride All in your Apache configuration."

## Some other notes:
My apache2.conf does have AllowOverride set to All.

js/default.js has been deleted in a recent change so now the redirect from js/canary.js to js/default.js doesn't work resulting in the error message and hard stop during installation.

Give us some context: 

* What database are you using? (e.g. mongo, mysql, postgres)
I'm using mysql.

esavir:
![known_issue](https://user-images.githubusercontent.com/49365518/56656296-60450c00-6652-11e9-892f-250475fd45c3.png)

esavir:
PR  #2410

PR #2410, opened 2019-04-24
Redirect to known.min.js in .htaccess

esavir:
The AllowOverride check tests that a redirect .htaccess works
Since js/default.js does not exist anymore we redirect to js/known.min.js

During install, there was a hardstop error that said the AllowOverride was not set to All in apache2.conf.

What actually happened was that the check was wrong because it was trying to redirect from js/canary.js to js/default.js and default.js did not exist anymore so that the error was incorrect.

Now, .htaccess redirects from js/canary.js to js/known.min.js so the AllowOverride check during install works.

## Here's why I did it:
The user can proceed with the install now.

## Checklist:

- [x] If this code includes interface changes, I've included screenshots in this Pull Request thread
screenshot of changed error message which now references known.min.js instead of default.js.
- [x] I can run the unit tests successfully.
got 7 failures from master branch

Error message with fixed file name.
known_issue_fixed

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -11,7 +11,7 @@
     RewriteCond $1::%{REQUEST_URI} ^(.*)::(.*?)\1$
     RewriteRule ^(.*)$ - [ENV=BASE:%2]
 
-    RewriteRule ^js/canary\.js$ %{ENV:BASE}/js/default.js [L]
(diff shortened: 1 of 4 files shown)


esavir:
PR for Fix #2409.

PR #2470, opened 2019-06-25
Associates object with a page when a permalink

benwerd:

Permalink pages now actually set the object into metadata.

## Here's why I did it:

I want to be able to reference the object itself outside of its display shell. This is an easy way to put the object somewhere the whole page, all templates, and other hooks can access it.

## Checklist:

### DIFF
--- a/Idno/Common/Page.php
+++ b/Idno/Common/Page.php
@@ -881,7 +881,7 @@ namespace Idno\Common {
          * @param bool $status Is this a permalink? Defaults to 'true'
          * @param object $entity Optionally, an entity this page is associated with
          */
-        function setPermalink($status = true, $entity = null)
+        function setPermalink(bool $status = true, Entity $entity = null)
         {
             $this->isPermalinkPage = $status;
             if ($status && $entity) $this->setEntity($entity);
--- a/Idno/Common/Page.php
+++ b/Idno/Common/Page.php
@@ -42,6 +42,10 @@ namespace Idno\Common {
         // to no, but you can use $this->setPermalink() to change this
         private $isPermalinkPage = false;
 
+        // If this page is associated with a particular entity, we'll
+        // store it here so we can access it in shell templates etc
+        private $entity = null;
+
         // Is this an XmlHTTPRequest (AJAX) call?
         public $xhr = false;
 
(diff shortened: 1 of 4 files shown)


Idno/Common/Page.php
```diff
@@ -847,14 +851,40 @@ function adminGatekeeper()
...
+
         /**
          * Is this page a permalink for an object? This should be set to 'true'
-         * if it is.
+         * if it is. Optionally, we can also associate the page with the object here.
          * @param bool $status Is this a permalink? Defaults to 'true'
+         * @param object $entity Optionally, an entity this page is associated with
          */
-        function setPermalink($status = true)
+        function setPermalink($status = true, $entity = null)
```
mapkyca: Consider using type hinting (as per style guidelines)

(this is relatively new guideline, so you may not be aware!)

Tests/Pages/User/ViewTest.php
```diff
@@ -18,12 +18,12 @@ function testWebmentionContent()
                 $notification = $eventdata['notification'];
             });
 
-            $source = 'http://karenpage.com/looking-for-information-' . md5(time() . rand(0, 9999));
+            $source = 'http://karenpage.dummy/looking-for-information-' . md5(time() . rand(0, 9999));
```
mapkyca: Consider using a dataProvider

Tests/Pages/User/ViewTest.php
```diff
@@ -8,16 +8,6 @@
     class ViewTest extends \Tests\KnownTestCase
     {
 
-        function testPagePermalink()
```
mapkyca: Why remove test? 

If the test is failing it shouldn't be removed, the root cause needs to be fixed.

mapkyca:
Binding a permalink to an object is definitely good.

Tests/Pages/User/ViewTest.php
benwerd: It was a test I added and then quickly removed because we don't have the capability right now to do what I needed it to. I'll work on the test infra separately and then bulk out the tests.

Idno/Common/Page.php
benwerd: Updated

Tests/Pages/User/ViewTest.php
benwerd: We're not actually iterating here, so I'm not sure there's call for one?

mapkyca: Yeah, my bad. Fighting through a cold!

PR #2471, opened 2019-06-25
Removing needless parameter

benwerd:

`getEntity` shouldn't take a parameter.

## Checklist:

### DIFF
--- a/Idno/Common/Page.php
+++ b/Idno/Common/Page.php
@@ -864,7 +864,7 @@ namespace Idno\Common {
          * @param $entity
          * @return object|null
          */
-        function getEntity($entity) {
+        function getEntity() {
             return $this->entity;
         }
 
--- a/Idno/Common/Page.php
+++ b/Idno/Common/Page.php
@@ -861,8 +861,7 @@ namespace Idno\Common {
 
         /**
          * Returns the entity associated with this page, if it exists
-         * @param $entity
-         * @return object|null
+         * @return ?Entity
          */
         function getEntity() {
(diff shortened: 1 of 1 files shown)


Idno/Common/Page.php
```diff
@@ -864,7 +864,7 @@ function setEntity($entity) {
          * @param $entity
```
mapkyca: Need to remove the code comment too

Idno/Common/Page.php
```diff
@@ -864,7 +864,7 @@ function setEntity($entity) {
...
          * @return object|null
          */
-        function getEntity($entity) {
+        function getEntity() {
```
mapkyca: If this is an object|null, consider return type hinting - e.g.  :  ? BaseObject

mapkyca: Docs are good, but also consider using the language construct, e.g.

```
use Idno\Common\Entity;

...

function getEntity($entity) : ? Entity {

...

}

```

benwerd: Done, although now I've done it, I don't really like how the file is now inconsistent. I'll make a note to go back and bring it all into line, but we should probably make a rule to keep files internally consistent. It would be inappropriate to do that in this PR, but that's something for another time.

benwerd: (Docs weren't good - phpdoc doesn't support this syntax. I've changed it back but added return type hinting.)

benwerd:
YOLO committing this one because it's tiny.

PR #2514, opened 2019-08-10
Adjust SQL

cdn:
Adjust SQL to reference the previously created table

Rewrote the date reference

## Here's why I did it:

SQL breaks otherwise


mapkyca:
Thanks!

PR #2637, opened 2019-12-14
Closes #2485: Splitting search into its own table

### LINKED ISSUE #2485: Datamodel: Get rid of search field
Having a munged search field as a column is a bad way of doing it. 

Not only is it non-relational, it's inefficient,  plus limits you to certain database engines on that table. 

The "correct" way to do this is to use Elastic Search or similar, but since that's not possible for most users, we should split the search fields out into their own tables

mapkyca:

Split search column into its own table

## Here's why I did it:

Slowly improving data model by splitting the full text search column into its own table. Later to be dispensed with for some proper search tool like elastic search

## Checklist:


mapkyca:
@benwerd  bump. Might just merge this, caveat emptor

Issue #2622, opened 2019-12-05
Remove dependence on jQuery


benwerd:
jQuery isn't necessary, and can present a security risk if it's not kept up to date.

* Remove jQuery
* Create simple library of internal DOM tools

PR #2638, opened 2019-12-16
Localise timeago substitutions

cdn:

Added in timeago locales file based on language->getLanguage()

 il y a environ 2 jours

## Here's why I did it:

Localisations are available and we aren't benefitting.

## Checklist:

- [/] This pull request addresses a single issue
- [?] If this code includes interface changes, I've included screenshots in this Pull Request thread
- [/] I've adhered to [Known's style guide](http://docs.withknown.com/en/latest/developers/standards/) ([these codesniffer rules](http://docs.withknown.com/en/latest/developers/testing/#code-style-testing) might help!)
- [/] I've tested my code in-browser

### DIFF
--- a/templates/default/shell/footerjavascript.tpl.php
+++ b/templates/default/shell/footerjavascript.tpl.php
@@ -17,10 +17,23 @@ if (\Idno\Core\Idno::site()->session()->isLoggedIn()) {
 
 }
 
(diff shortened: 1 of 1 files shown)


cdn:
There is an alternative library; https://github.com/hustcc/timeago.js - have been exploring using this instead, which would be one step towards removing jquery #2622

PR #2652, opened 2020-01-08
Localisation of bootstrap toggle controls

cdn:

Made `data-on` and `data-off` localisable

## Checklist:

- [?] This pull request addresses a single issue
- [/] I've tested my code in-browser


languages/source/known.pot
```diff
@@ -51,7 +51,7 @@ msgid "Missing X-Known-Service-Signature, service call is not possible."
 msgstr ""
 
 #: ./Idno/Core/Service.php:27
-msgid "Sorry, signature doesn\'t match up."
```
mapkyca: How are these strings generated? If there is a problem with the .pot generator file I'd love a patch for that, as otherwise these changes will get clobbered in a future build

cdn: Where a single-quoted string in PHP needs the prime apostrophes escaped, the po[t] file doesn't.

Additionally the quotes in html in the `msgid`s get over-escaped.

The Aleph strings are particularly awkward, and somewhat redundant : )

mapkyca: Hmm... I'll see if I can get time to look at that. https://github.com/mapkyca/known-language-tools will need updating by the sounds of it.

cdn: https://github.com/cdn/Known/blob/localisation/templates/default/robot/wizard.tpl.php reduces the variations to: three prefix words and the main body about sharing to Fb/T; known-language-tools as-is fails on that part; will have to exercise the regex part of my brain :o)

PR #2653, opened 2020-01-09
Rework localisation strings so they do not trip language tools…

cdn:
… buildpot.php

Wizard and one other

Easier than sorting out the regex issue(s)


cdn:
There’s also Line 168 in `./Idno/Pages/Admin/Users.php`.
That gets prematurely truncated by the `known-language-tools` regex.

Fixes include:

> "Usernames can't contain a slash (\&#8216;/\&#8217;) character."

and

> "Usernames can't contain a slash %s character.", ["('\')"]

It depends on how strongly people feel about quote types.

PR #2687, opened 2020-02-11
Use ENV as source of Install Config

Lewiscowles1986:

Environment aware base-config

## Here's why I did it:

I have setup Known on Heroku, the ephemeral filesystem made installing a royal pain. The same would be true of CloudFoundry.

## Checklist:

- [X] This pull request addresses a single issue
- [X] If this code includes interface changes, I've included screenshots in this Pull Request thread
- [X] I've adhered to [Known's style guide](http://docs.withknown.com/en/latest/developers/standards/) ([these codesniffer rules](http://docs.withknown.com/en/latest/developers/testing/#code-style-testing) might help!)
- [X] My git branch is named in a descriptive way - i.e., yourname-summary-of-issue
- [X] I've tested my code in-browser
- [X] My code contains descriptive comments
- [X] I've added tests where applicable, and...
- [X] I can run the [unit tests](http://docs.withknown.com/en/latest/developers/testing/#unit-testing) successfully.

### DIFF
--- a/Idno/Core/Config.php
+++ b/Idno/Core/Config.php
@@ -182,6 +182,50 @@ namespace Idno\Core {
                             }
                         }
                     }
+                    if (array_key_exists('KNOWN_DATABASE_URL', $_ENV)) {
+                        $parsed = parse_url($_ENV['KNOWN_DATABASE_URL']);
+                        $this->ini_config['database'] = $parsed['scheme'];
+                        $this->ini_config['dbname'] = basename($parsed['path']);
+                        $this->ini_config['dbuser'] = $parsed['user'];
+                        $this->ini_config['dbpass'] = $parsed['pass'];
+                        $this->ini_config['dbhost'] = $parsed['host'];
+                    }
+
+                    $cloudcube = array_key_exists('CLOUDCUBE_URL', $_ENV);
+                    $aws_s3 = array_key_exists('KNOWN_AWS_S3_BUCKET', $_ENV);
+                    $path = '/Uploads';
+                    $bucket = '';
+
+                    if ($cloudcube) {
+                        $parsed = parse_url($_ENV['CLOUDCUBE_URL']);
+                        $host_parts = explode('.', $parsed['host']);
+                        $bucket = $host_parts[0];
+                        $this->ini_config['aws_key'] = $_ENV['CLOUDCUBE_ACCESS_KEY_ID'];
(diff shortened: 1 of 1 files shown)


Idno/Core/Config.php
```diff
@@ -182,6 +182,50 @@ function loadIniFiles()
                             }
                         }
                     }
+                    if (array_key_exists('DATABASE_URL', $_ENV)) {
```
mapkyca: Hmm... 

If we're going to allow $_ENV variables, which I'm not entirely opposed to, I'd prefer it applied in a slightly more universal way so that we can pass other variables as well. 

I imagine using the KNOWN_ prefix like we already do with $_SERVER...

Lewiscowles1986: > If we're going to allow $_ENV variables, which I'm not entirely opposed to, I'd prefer it applied in a slightly more universal way so that we can pass other variables as well.

Can you put this another way. I'm not entirely sure what you mean by it.

> I imagine using the KNOWN_ prefix like we already do with $_SERVER...

Is this related to the above? I can absolutely prefix all the ENV, although for some addons it will mean a less straightforward user-experience.

The specific cloud experience this is working towards is a one-click heroku installer. It'll be less turn-key, and more error prone if users have to manually change things

Lewiscowles1986: It could also benefit docker if that is something you're interested in. I'm trying to keep the shared-hosting ability of this as much as possible.

Lewiscowles1986: @mapkyca this now pulls everything which can be aliased including a prefix  `KNOWN_`

`CLOUDCUBE` has to be an exception as it does not support deploy-time or automated aliasing sadly.

mapkyca: Was thinking more that we'd strip the KNOWN_ off the ENV var so that we can use the same mechanism to set other variables

Lewiscowles1986: Come again? Above it seems you wanted to prefix. Do you have a more complete vision of what you need to get this into known?

Lewiscowles1986:
Thanks @benwerd please let me know if changes are needed, I'm happy to try to maintain things and not drive-by commit

benwerd:
I think this looks great. If we need to refactor the way this is done, we’ll figure that out - but that’s not in the scope of this PR.

PR #2698, opened 2020-03-02
Add comments to .htaccess about upload size workaround

lindner:

I had mysterious problems with file uploads, figured it out and the default 8M php upload size limits was causing problems.

## Here's why I did it:

Decided to add a commented entry in the .htaccess file.  Should also probably add this to the diagnostics page?  Maybe add to FAQ as well?  At least this is a start....

## Checklist:

- [ ] This pull request addresses a single issue

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -97,4 +97,9 @@ AddType video/ogg .ogv
 AddType video/mp4 .mp4
 AddType video/webm .webm
 
+# If you have problems uploading large images try uncommenting these lines
+#
+#   php_value post_max_size 64M
+#   php_value upload_max_filesize 60M
+
 ## END Known Webserver Config (don't remove)


mapkyca:
TBH I'm tempted to remove everything except the rewrite stuff from the htaccess - my feeling is that this is more of a platform config thing rather than doing hacky things in htaccess.

lindner:
Sure, but we should address the upload size issue somewhere,
especially because it's a php platform thing...

Maybe warn during install/warmup if phpinfo shows small values?

benwerd:
There are unfortunately quite a few shared hosts who still _only_ allow you to change this in the htaccess, at least easily. I think the config stuff has to stay.

mapkyca:
Hmmmm... maybe. My experience is that software with weird things in their .htaccess tends to fight with the server wide config which should be the canonical one. Generally people set stuff in the .htaccess and then the devs forget to update it, time moves on and you find that things don't work because the software's settings - generous at the time - are very conservative in $currentyear.

Wordpress for example only has rewrite rules in their distributed .htaccess, and I think this is the model we should follow.

But have it your way, this isn't the hill I want to die on.

PR #2721, opened 2020-04-02
MySQL warmup schema edits

Lewiscowles1986:

I altered the MySQL warmup schema to be compatible with a wider number of MySQL server backends

## Here's why I did it:

Heroku MySQL does not support FULLTEXT indexes using clearDB free tier. There seems to be sparse documentation other than the SQL failing on creation of the fulltext index.

I have not worked out how to add back in for users.

- [X] This pull request addresses a single issue
- [X] If this code includes interface changes, I've included screenshots in this Pull Request thread
- [X] I've adhered to [Known's style guide](http://docs.withknown.com/en/latest/developers/standards/) ([these codesniffer rules](http://docs.withknown.com/en/latest/developers/testing/#code-style-testing) might help!)
- [X] My git branch is named in a descriptive way - i.e., yourname-summary-of-issue
- [X] I've tested my code in-browser
- [X] I can run the [unit tests](http://docs.withknown.com/en/latest/developers/testing/#unit-testing) successfully.

### DIFF
--- a/warmup/schemas/mysql/2014100801.sql
+++ b/warmup/schemas/mysql/2014100801.sql
@@ -19,8 +19,7 @@ CREATE TABLE IF NOT EXISTS `reader` (
   PRIMARY KEY (`uuid`),
   UNIQUE KEY `_id` (`_id`),
   KEY `owner` (`owner`,`created`),
-  KEY `entity_subtype` (`entity_subtype`),
-  FULLTEXT KEY `search` (`search`)
+  KEY `entity_subtype` (`entity_subtype`)
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8;
 
 UPDATE `versions` SET `value` = '2014100801' WHERE `label` = 'schema';
--- a/warmup/schemas/mysql/2019121401.sql
+++ b/warmup/schemas/mysql/2019121401.sql
@@ -2,8 +2,7 @@
 CREATE TABLE IF NOT EXISTS `entities_search` (
   `_id` varchar(32) NOT NULL,
   `search` longtext NOT NULL,
-  PRIMARY KEY (`_id`),
-  FULLTEXT KEY `search` (`search`)
+  PRIMARY KEY (`_id`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
 INSERT INTO `entities_search` SELECT `_id`, `search` FROM `entities`;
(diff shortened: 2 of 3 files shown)


mapkyca:
Hmm...

As it stands, this will break search for longer strings. IIRC LIKE is used for search terms < 5 characters, and for anything longer the full text index is used. So, at the very least there also needs to be a change to the MySQL.php data handler.

I would prefer it if we could keep the full text search, and maybe make it an optional for Heroku somehow. This in the short term.

That being said, dropping full text search long term does make sense. Doing search in the database is Bad. The correct solution is to use Lucene / ElasticSearch (which I've done several times before).

Lewiscowles1986:
> I would prefer it if we could keep the full text search, and maybe make it an optional for Heroku somehow. This in the short term.

I Agree. I'll look into the data handler.

RE: Lucene / ElasticSearch, I'm not sure I'd like to force people into that.

Postgres doesn't use the full text search does it? I didn't notice any special purpose indexes.

Is it that each database provider i.e. MySQL has it's own rules? I didn't look too deeply into this. I'll iterate.

mapkyca:
> RE: Lucene / ElasticSearch, I'm not sure I'd like to force people into that.

Not forced, no. I imagine having a basic search functionality implemented in database, but with hooks for plugins that can easily replace this. Elastic would be implemented as a plugin, fairly easy to do and I've written it several times before (but all those plugins are unfortunately proprietary)

> Postgres doesn't use the full text search does it? I didn't notice any special purpose indexes.

IIRC it does, but I don't think it's explicitly declared in the schema. I don't really know Postgres, so I've kind of left that for other folks to take care of!

> Is it that each database provider i.e. MySQL has it's own rules? I didn't look too deeply into this. 

Yes, there's an inheritance model, but there are specific implementations of certain methods for each engine.

Something I'd like to do if I had copious amounts of free time would be to re-implement this using something like doctrine dbal.

Lewiscowles1986:
It's taken me this long to locate. I think the only place it uses fulltext search is this block

https://github.com/idno/known/blob/master/Idno/Data/MySQL.php#L590

I think this means that the config has a capacity to disable full-text search; so the schema / migrations could use that same hook, and have a migration for full-text if that is set to false.

Then I can set heroku deploys to require that for MySQL and everything uses existing ingredients more-or less and I think should avoid problems.

mapkyca:
That would be a way around it, certainly. I think that's the only place it's looked for. There's a similar shortcut in the Postgres handler too, if memory serves. 

There would need to be some tweaks to the warmup sql loader for this to work, as at the moment it just goes through the whole file. So that's probably splitting the file, and then modifying both the web and cli installer.

Issue #2451, opened 2019-06-10
MySQL should use utf8mb4 collation


mapkyca:
## While trying to do this:

Store any character not in the BMP - emojis, mathematical symbols, etc on mysql

## I encountered this error:

They can't be stored

## Some other notes:

It is recommended widely (since ~2016 at least) that you don't use utf8 (which is an alias for utf8mb3) and instead use utf8mb4.

Postgres is unaffected, as 4byte utf is the default

Give us some context: 

* It'd also be really handy if you could tell us the contents of your ```version.known``` file
* What database are you using? (e.g. mongo, mysql, postgres)
* Any warnings or errors in your admin/diagnostics page?
* If this is a programming bug, can you include examples of any Micropub / API calls / webhook pings you make? Otherwise please don't worry about what this means!
* Bonus points - are you able to illustrate the issue with a unit test? If so, submit it as a pull request!

mapkyca:
Low priority, possibly, since object serialisation seems to handle this ok. Will be an issue for any raw body fields and metadata

lindner:
Hitting this now...   I can't believe they screwed up charsets this badly....  Wish I could use postgres instead of this ugliness...

lindner:
Some details here:

https://mathiasbynens.be/notes/mysql-utf8mb4

lindner:
definitely affects the search tables.

mapkyca:
Migration for this is actually a little fiddly owing to mysql's key length restrictions. Not had the bandwidth to look at it.

I'm tempted to address this when we move towards changing the data model around, and also thinking about binning off full text search altogether in favour of an Elastic Search / Lucine plugin. Search really isn't something to be doing at the database level.

thierrymarianne:
Hello @lindner, @mapkyca,

For your information, altering tables was not enough for me. 

I had this MySQL 5.7 database which still left me in the dark with non-terminating queries. 

Upgrading to MySQL 8.0 apparently did the trick. 

I could find the queries for which emojies turned out to be rejected 
by v5.7 MySQL database for running tests if that could help.

lindner:
5.7 is EOL in one year.  I think we can bump to 8.0 as minimum for this version unless there's an active user that really wants to stick with that...

[Same goes for other infrastructure pieces, we can and should specify a modern PHP etc...]

benwerd:
+1 from me, fwiw!

PR #2737, opened 2020-04-21
Schema: Split metadata table out into separate collection based tables

mapkyca:

Splits metadata tables out into their own tables.

## Here's why I did it:

This is the onward goal of improving the database schema bit by bit.

Splitting metadata out into their own tables should make things a bit faster. 

More importantly makes the schema easier to work with, for example, we can offload data integrity (e.g. deleting metadata) off to the database layer to manage. 

It also means we might be able to move over utf8mb4 at some point in the future (#2451), for which the length of indexes proved blockers.

### DIFF
--- a/Idno/Data/MySQL.php
+++ b/Idno/Data/MySQL.php
@@ -75,6 +75,7 @@ namespace Idno\Data {
                             2017032001,
                             2019060501,
                             2019121401,
+                            2020042101,
                         ] as $date) {
                             if ($basedate < $date) {
                                 if ($sql = @file_get_contents($schema_dir . $date . '.sql')) {
@@ -110,7 +111,14 @@ namespace Idno\Data {
         {
             try {
                 $this->client->query("optimize table entities");
-                $this->client->query("optimize table metadata");
+                $this->client->query("optimize table entities_metadata");
+                $this->client->query("optimize table entities_search");
+                $this->client->query("optimize table config");
+                $this->client->query("optimize table config_metadata");
+                $this->client->query("optimize table config_search");
+                $this->client->query("optimize table reader");
+                $this->client->query("optimize table reader_metadata");
+                $this->client->query("optimize table reader_search");
                 $this->client->query("optimize table session");
             } catch (\Exception $e) {
                 error_log($e->getMessage());
@@ -251,7 +259,7 @@ namespace Idno\Data {
                     
                     
                     $retval = $array['_id'];
-                    if ($statement = $client->prepare("delete from metadata where _id = :id")) {
+                    if ($statement = $client->prepare("delete from {$collection}_metadata where _id = :id")) {
                         $statement->execute(array(':id' => $array['_id']));
                     }
 
@@ -277,8 +285,8 @@ namespace Idno\Data {
                             if (strlen($value) > 255) { // We only need to store the first 255 characters
                                 $value = substr($value, 0, 255);
                             }
-                            if ($statement = $client->prepare("insert into metadata set `collection` = :collection, `entity` = :uuid, `_id` = :id, `name` = :name, `value` = :value")) {
-                                $statement->execute(array('collection' => $collection, ':uuid' => $array['uuid'], ':id' => $array['_id'], ':name' => $key, ':value' => $value));
+                            if ($statement = $client->prepare("insert into {$collection}_metadata set `_id` = :id, `name` = :name, `value` = :value")) {
+                                $statement->execute(array(':id' => $array['_id'], ':name' => $key, ':value' => $value));
                             }
                         }
                     }
@@ -396,7 +404,7 @@ namespace Idno\Data {
                 $offset           = (int)$offset;
                 $where            = $this->build_where_from_array($parameters, $variables, $metadata_joins, $non_md_variables, 'and', $collection);
                 for ($i = 1; $i <= $metadata_joins; $i++) {
-                    $query .= " left join metadata md{$i} on md{$i}.entity = {$collection}.uuid ";
+                    $query .= " left join {$collection}_metadata md{$i} on md{$i}._id = {$collection}._id ";
                 }
                 if (isset($parameters['$search'])) {
                     $query .= " left join {$collection}_search srch on srch._id = {$collection}._id ";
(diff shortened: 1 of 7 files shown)


Lewiscowles1986:
This isn't working when rebased against an existing deploy using postgres for heroku. Specifically it does not find the sole existing user, which it's been confirmed at the database level still exists.

Lewiscowles1986:
These two fixed being able to see my site, but not logging in

```
INSERT INTO config_metadata
SELECT _id, name, value FROM metadata WHERE collection = 'config'
ON CONFLICT DO NOTHING;
```

```
INSERT INTO entities_metadata
SELECT _id, name, value FROM metadata WHERE collection = 'entities'
ON CONFLICT DO NOTHING;
```

there is an interesting property in that the name and _id do not seem to be unique constraint enforced

mapkyca:
Yeah, it would actually help if I _executed_ the patch script. Serves me for doing late night coding. Try now.

(obviously back up first)

mapkyca:
Regarding unique. Yes, this is known and it is Known. Saving an entity clears all metadata for that entity and resets. Not the _best_ way, just _a_ way.

Handling metadata more efficiently is definitely on the list, but I'm not there yet.

Lewiscowles1986:
```sql
TRUNCATE TABLE config_metadata;
ALTER TABLE config_metadata ADD UNIQUE(_id, name);

INSERT INTO config_metadata
SELECT _id, name, value FROM metadata WHERE collection = 'config'
ON CONFLICT DO NOTHING;

TRUNCATE TABLE entities_metadata;
ALTER TABLE entities_metadata ADD UNIQUE(_id, name);

INSERT INTO entities_metadata
SELECT _id, name, value FROM metadata WHERE collection = 'entities'
ON CONFLICT DO NOTHING;
```

This is my just-now done fix. It does mean I'll need to run some checks to ensure it's not run per-deploy.

Idno/Data/MySQL.php
```diff
@@ -75,6 +75,7 @@ function checkAndUpgradeSchema()
                             2017032001,
                             2019060501,
                             2019121401,
+                            2020042101,
```
Lewiscowles1986, line 78: is this one supposed to update too?

mapkyca: Only if you're running with the mysql engine. 

Scripts only execute for the engine you're on.

warmup/schemas/mysql/mysql.sql
```diff
@@ -29,6 +29,16 @@ CREATE TABLE IF NOT EXISTS `config_search` (
...
 
+CREATE TABLE IF NOT EXISTS `config_metadata` (
+  `_id` varchar(32) NOT NULL,
+  `name` varchar(64) NOT NULL,
+  `value` text NOT NULL,
+  KEY `value` (`value`(255)),
+  KEY `name` (`name`),
+  KEY (`_id`),
+  CONSTRAINT `cm_id_id` FOREIGN KEY (`_id`) REFERENCES `config` (`_id`) ON DELETE CASCADE ON UPDATE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
Lewiscowles1986, line 40: UNIQUE constraint could go here. Basically in a fairly new (April 1st) Install, I had duplicate values stored which were from the application when running the `INSERT INTO SELECT FROM WHERE` part of the queries

warmup/schemas/postgres/2020042101.sql
```diff
@@ -0,0 +1,63 @@
...
+INSERT INTO config_metadata SELECT _id, name, value FROM metadata WHERE collection = 'config';
+
+
+
+CREATE TABLE IF NOT EXISTS reader_metadata (
+  _id varchar(32) NOT NULL,
+  name varchar(64) NOT NULL,
+  value text NOT NULL,
+  FOREIGN KEY (_id) REFERENCES reader (_id) ON DELETE CASCADE ON UPDATE CASCADE
+);
```
Lewiscowles1986, line 37: UNIQUE constraint could go here. Basically in a fairly new (April 1st) Install, I had duplicate values stored which were from the application when running the INSERT INTO SELECT FROM WHERE part of the queries

warmup/schemas/postgres/postgres.sql
```diff
@@ -28,6 +28,17 @@ CREATE TABLE IF NOT EXISTS config_search (
   FOREIGN KEY (_id) REFERENCES config (_id) ON DELETE CASCADE ON UPDATE CASCADE
 );
 
+CREATE TABLE IF NOT EXISTS config_metadata (
+  _id varchar(32) NOT NULL,
+  name varchar(64) NOT NULL,
+  value text NOT NULL,
+  FOREIGN KEY (_id) REFERENCES config (_id) ON DELETE CASCADE ON UPDATE CASCADE
```
Lewiscowles1986, line 35: I think both warmup and patch need the unique constraint added. It is quite confusing both exist to be honest

warmup/schemas/postgres/2020042101.sql
mapkyca: Have to check how this affects the storage of array values.

Can't remember whether arrays are serialised in metadata or stored multiple times (I think the latter). Switched between projects so many times I can't remember off the top of my head

warmup/schemas/postgres/postgres.sql
mapkyca: Warmup is canonical, patch is to roll up existing users as defined by the schema version number

Idno/Data/MySQL.php
Lewiscowles1986: Right. Well the postgres edits are not in-line with the MySQL ones then. It's one of the parts that was confusing.

warmup/schemas/postgres/postgres.sql
Lewiscowles1986: so it does protect against a patch running a second time?

mapkyca: In a crude way. Patch files update the schema version as they go.

Something I want to do, but probably will never have time to do, is move this stuff all over to doctrine dbal and have schema definitions stored in json so they can be managed automatically.

Done this for another project and it removes a lot of headaches.

Idno/Data/MySQL.php
mapkyca: Well, some patches are only relevant to specific database engines hence why the difference (also, Postgres isn't so much a well trodden path)

PR #2747, opened 2020-04-28
Update Postgres upgrade script to work

### LINKED ISSUE #2746: Postgres; Patching for split metadata tables will fail on inconsistent metadata
## While trying to do this:

When running the patch script https://github.com/idno/known/blob/master/warmup/schemas/postgres/2020042101.sql introduced by 56c761b17cc6ed6f1658752424f7f6f4257e857b

## I encountered this error:

This will fail if the metadata table contains metadata that references entities that don't exist. This can happen only in cases where something has gone wrong with object deletion at some point, and so shouldn't be widespread (this issue is part of the reason why I made this patch).

However the issue was present in an old site I rolled up last night. I fixed it for mysql (c6b887f9e9da167acddfa8e2f3776a12283298a4) by temporarily disabling the foreign key checks on the upgrade script. 

A similar fix needs to be applied to Postgres. 

## Some other notes:

I don't think doing this in Postgres is quite as simple, and I believe you need to do some sleight of hand disabling triggers on tables and re-enabling them later. Not had time to dig in to it in much detail.

/cc @Lewiscowles1986

Lewiscowles1986:
Fixes #2746

Ignores deprecated_metadata

It is unknown / unknowable at this point if the constraint for unique _id, name pairs should be altered into this script, so that if the script errors but allows metadata to drop after transition to deprecated_metadata, if another script could be run without creating duplicates...

### DIFF
--- a/warmup/schemas/postgres/2020042101.sql
+++ b/warmup/schemas/postgres/2020042101.sql
@@ -10,9 +10,9 @@ CREATE INDEX CONCURRENTLY IF NOT EXISTS m_value ON entities_metadata (value);
 CREATE INDEX CONCURRENTLY IF NOT EXISTS m_name ON entities_metadata (name);
 CREATE INDEX CONCURRENTLY IF NOT EXISTS m__id ON entities_metadata (_id);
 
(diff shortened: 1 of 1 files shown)


mapkyca:
Thanks!

PR #2756, opened 2020-05-02
Removing home rolled JWT class.

mapkyca:
Removing home rolled JWT class.
## Here's why I did it:
We shouldn't be re-inventing the wheel, and instead should use one of the many libraries to handle this


Lewiscowles1986:
Shouldn't this try to provide a compatible API while using an external JWT? I only see a removal, which might mean breakages.

mapkyca:
Should be fine - AFAIK I was the only one who was using it, and it was somewhat broken in any case. The libraries that use this (all the oauth stuff) has already been refactored.

PR #2884, opened 2021-02-03
Forwarding to login now operates through a consistent method

benwerd:

I added a `$Page->forwardToLogin($fwd);` method.

## Here's why I did it:

We were reproducing this code all over the place. Some plugins were also implementing it, but because the method has changed, none of them forward. Adding a wrapper function will streamline this process, make life easier for developers, and help ensure that plugins don't randomly break when  we change the methodology in  the future.


lindner (commented):
LGTM

Consider adding Issues to migrate all existing code to use these constructs?

PR #2912, opened 2021-03-14
Adding Codesniffer to default distro

benwerd:

Added PHP-Codesniffer library and config to default distribution.

## Here's why I did it:

It doesn't make much sense to have this sit in a separate repo, and the repo was referencing a version that was out of date.

### DIFF
--- a/.phpcs.xml
+++ b/.phpcs.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0"?>
+<ruleset name="known">
+    <description>The Known coding standard.</description>
+    <arg name="tab-width" value="4"/>
+    <rule ref="Zend">
+        <exclude name="Generic.Files.LineLength.TooLong"/>
+        <exclude name="Generic.Files.LineLength.MaxExceeded"/>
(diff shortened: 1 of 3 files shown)


mapkyca:
Why remove the Known PHP CS package (which also has a place for the Known config files)?

benwerd:
A couple of reasons, but the main one is that the config file can then sit in the root, allowing us to leave it unspecified on the command line. The repo wasn't doing much, and it seems to me that something required to develop on the codebase should be referenced directly rather than through a wrapper.

mapkyca:
Hmm.. fair enough but I was trying to remove the cruft that was building up in the root

lindner:
I do think that having this in a separate repo is important for consistency across this repo and the plugins.  But there also needs to be a lot that needs to happen to get all the plugins updated to the latest.  ex:

- CI coverage
- rename master to main
- etc.

benwerd:
I'm going to close this and take a different approach. The point about plugins is well-taken.

PR #2921, opened 2021-04-14
Adding web monetization

benwerd:

Users can now add a web monetization pointer

## Here's why I did it:

We should be promoting distributed monetization.


lindner:
Does this always display on the home page or is it hidden?

benwerd:
It's an option in the admin panel - there's no visible component.

mapkyca:
Ahh... is Coil something new? (terrible name)

I wondered why my XRP position suddenly spiked.

mapkyca:
Would be nice to accept BTC directly. IIRC some exchanges e.g blockchain have a payment API, or at least did way back when.

benwerd:
It's not _very_ new, but it's gaining steam. See https://coil.com for more - and https://www.grantfortheweb.org/

benwerd:
I wrote the Unlock plugin to allow Known site users to receive ETH directly - and obvs other coins are available. BTC isn't that suitable for site-to-site payments IMO because it's relatively slow and inflexible.

mapkyca:
It's slower than a credit card, true, but so much faster than dealing with banks for overseas payments. These days half my clients pay me in crypto, especially in those places which have "issues" with money - i.e. China.

mapkyca:
Different use cases obviously. Big into crypto these days :D

benwerd:
Same - or rather, decentralization. But let's take the discussion off this PR and put it somewhere else ;)

mapkyca:
Heh heh! You mean this isn't IRC?

Cambridgeport90:
Good call... except it will be even cooler when more cryptocurrencies are supported. Hive is my favourite... because it was designed from the ground up to be for bloggers and creators.

PR #2932, opened 2021-04-29
Check for edit permissions before save

benwerd:

The `Entity::save()` method now, by default, checks that the current user can edit the entity before saving to the database. 

**Please note:** this will currently fail tests, until OAuth plugin changes are merged.

## Here's why I did it:

This takes responsibility for this kind of check from plugins, which have not always be doing it effectively.

### DIFF
--- a/Idno/Entities/User.php
+++ b/Idno/Entities/User.php
@@ -610,7 +610,7 @@ namespace Idno\Entities {
                     }
 
                     $acl->addMember($user->getUUID());
-                    $acl->save();
+                    $acl->save(true);
 
                     \Idno\Core\Idno::site()->events()->triggerEvent('follow', array('user' => $this, 'following' => $user));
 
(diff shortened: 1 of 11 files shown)


Idno/Entities/User.php
```diff
@@ -610,7 +610,7 @@ function addFollowing($user)
                     }
 
                     $acl->addMember($user->getUUID());
-                    $acl->save();
+                    $acl->save(true);
```
lindner, line 613: small preference for something like `$acl->save(overrideAccess: true)` to make this more readable.  (for all usages)

lindner (approved):
LGTM, but should wait for test fixes and optional readability concerns before merge.

Cambridgeport90:
Makes perfect sense to me.

benwerd:
@mapkyca thanks for merging the first oauth change - just waiting on https://github.com/mapkyca/known-oauth2-client/pull/6

benwerd:
Btw @lindner named parameters are only supported in PHP 8 - so I don't think we can include them in Known core just yet. Most IDEs will identify the parameter name in the interface, but I agree that it's not hugely legible.

Issue #3002, opened 2021-10-08
MySQL Schema - Where is it?


teamayu:
In your documentation instructions for installation it says:

"For now, you will need to load the SQL schema stored in /schemas/mysql/mysql.sql."

Where is this mysterious 'schemas' directory? 

Thank you.

cdn:
https://github.com/idno/known/tree/dev/warmup, so; `/warmup/schemas/mysql/mysql.sql`

teamayu:
Ah, got it. It's in the warmup folder. You might want to update the docs, thank you.

benwerd:
Fixed the instructions. Sorry for the confusion and thanks for the heads up!

PR #3003, opened 2021-10-09
Fixing schema path in instructions

benwerd:

I fixed the instructions to contain the correct path to the schema.

## Here's why I did it:

It was wrong and confusing people - see #3002.


lindner (commented):
LGTM

PR #3065, opened 2022-04-12
Improved malicious file filtering

benwerd:

Improved file matching.

## Here's why I did it:

There were a bunch of false positives related to animated GIFs.


lindner (approved):
LGTM

PR #3162, opened 2023-11-10
Wrap Checkin.tpl.php map JS for local variables

### LINKED ISSUE #3161: Hyphen in map div id prevents display of tiles
## While trying to do this:
See a map after a checkin. This URL for example https://stream.jeremycherfas.net/2023/checked-into-lombra-del-leone

## I encountered this error:
The map does not display

## Some other notes:
I looked in the console, and saw an error: `"Uncaught SyntaxError: Unexpected token '-'", filename: 'https://stream.jeremycherfas.net/2023/checked-into-lombra-del-leone', lineno: 301, colno: 28, …}`

That line contains `var map6fe9a186-a0ed-4660-b8c3-e7e0b52d723c = L.map('map_6fe9a186-a0ed-4660-b8c3-e7e0b52d723c', {`

If I simplify the map div's id, from `<div id="map_6fe9a186-a0ed-4660-b8c3-e7e0b52d723c"` to `<div id="map_` and make the same changes in the script the map does then display.

It still gives an error with regard to the use of Stamen tiles, but that is a different issue, which it will be much easier to try and solve when I can be sure that the map will display.

I suppose the function that creates the map div's id needs to be tweaked, but I do not know how to do that.

Give us some context: 

* It'd also be really handy if you could tell us the contents of your ```version.known``` file 
version = '1.6'
build = 2023102301
* What database are you using? (e.g. mongo, mysql, postgres)
mysql

* Any warnings or errors in your admin/diagnostics page?
See above
 
* If this is a programming bug, can you include examples of any Micropub / API calls / webhook pings you make? Otherwise please don't worry about what this means!
* Bonus points - are you able to illustrate the issue with a unit test? If so, submit it as a pull request!

Zegnat:

Fixes #3161 by removing random strings from JavaScript variable names. Instead the whole script is wrapped in [an IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) so short variable names can be used without conflicting in the global scope.

## Here's why I did it:

The `$object->_id` code was adding `-` (minus-hyphen) symbols within JavaScript variable names, where they are not allowed. This meant that maps could no longer be rendered on check-in type posts.

## Checklist:

### DIFF
--- a/IdnoPlugins/Checkin/templates/default/entity/Checkin.tpl.php
+++ b/IdnoPlugins/Checkin/templates/default/entity/Checkin.tpl.php
@@ -39,17 +39,18 @@
 </div>
 <?php if (empty($vars['feed_view'])) { ?>
         <script>
-            var map<?php echo $object->_id?> = L.map('map_<?php echo $object->_id?>', {
-                touchZoom: false,
-                scrollWheelZoom: false
-            }).setView([<?php echo $object->lat() ?>, <?php echo $object->long() ?>], 16);
-            var layer<?php echo $object->_id?> = new L.StamenTileLayer("toner-lite");
-            map<?php echo $object->_id?>.addLayer(layer<?php echo $object->_id?>);
-            var marker<?php echo $object->_id?> = L.marker([<?php echo $object->lat()?>, <?php echo $object->long()?>]);
-            marker<?php echo $object->_id?>.addTo(map<?php echo $object->_id?>);
-            //map<?php echo $object->_id?>.zoomControl.disable();
-            map<?php echo $object->_id?>.scrollWheelZoom.disable();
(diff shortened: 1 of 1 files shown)


Zegnat:
@jeremycherfas would you be able to test this new [Checkin.tpl.php](https://github.com/idno/known/blob/98a78dbb73a66dabe7b95e854ac9edb7e3ff54a3/IdnoPlugins/Checkin/templates/default/entity/Checkin.tpl.php) file and confirm that the issue goes away?

jeremycherfas:
I can confirm that this change fixes my problem with the display of maps.

I still need to work on a new tile provider, but at least I will know where those problems lie.

benwerd (approved):
Thank you!!

Issue #2461, opened 2019-06-16
Datamodel: Annotations are stored wrong


mapkyca:
## While trying to do this:

List, edit, delete or otherwise manage annotations and comments

## I encountered this error:

It's not really possible, as annotations are stored directly on the object

## Some other notes:

The current data model is very heavily influenced by the fact it was originally stored in mongo only. Storing annotations on the actual object is utterly crazy when you're in a relational world.

Redesign the data model to store annotations in its own table.

* [ ] Annotations to have its own relational table
* [ ] Object methods to address this for SQL stores
* [ ] Migration script for existing annotations

Give us some context: 

* It'd also be really handy if you could tell us the contents of your ```version.known``` file
* What database are you using? (e.g. mongo, mysql, postgres)
* Any warnings or errors in your admin/diagnostics page?
* If this is a programming bug, can you include examples of any Micropub / API calls / webhook pings you make? Otherwise please don't worry about what this means!
* Bonus points - are you able to illustrate the issue with a unit test? If so, submit it as a pull request!

mediaformat:
> Storing annotations on the actual object...

I'm encountering this very issue while trying to fetch individual reply annotations, the whole object is returned... 
#3186

mediaformat:
While we are refactoring relations between tables, I would also make annotations have a relation to their  (remote) user entity.

This will ensure remote user info such as `owner_name` & `owner_image` are always current, for example.

PR #3216, opened 2024-04-04
ActivityPub: Sprint 2

mediaformat:
# ActivityPub: Sprint 2 

- [x] Setup / Planning
- [ ] Migrate annotations
- [ ] Adapt/update existing annotations db querying code
- [ ] Minimally implement `http-foundation` Request methods for use with `landrok/activitypub`

Other
- [x] fix: adds $wbr to `parseURLs` closure

## Here's why I did it:
- Known's [Datamodel needs optimizing](https://github.com/idno/known/issues/2461) before the floodgates of federated activities are opened!

## Checklist:

- [ ] This pull request addresses a single issue


mediaformat:
@benwerd The following describes the Annotations schema, and what I've considered:
``` SQL
CREATE TABLE IF NOT EXISTS `annotations` (
  `uuid` varchar(255) NOT NULL,
  `_id` varchar(36) NOT NULL,
  `siteid` varchar(36),
  `owner` varchar(255) NOT NULL,
  `entity_subtype` varchar(64) NOT NULL,
  `created` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
  `contents` longblob NOT NULL,
  `publish_status` varchar(255) NOT NULL DEFAULT 'published',
  `entity_id` varchar(255) NOT NULL,
  `parent_id` varchar(255),
  PRIMARY KEY (`uuid`),
  UNIQUE KEY `_id` (`_id`),
  KEY `siteid` (`siteid`),
  FOREIGN KEY (owner) REFERENCES entities(uuid),
  KEY `entity_subtype` (`entity_subtype`),
  KEY `publish_status` (`publish_status`),
  FOREIGN KEY (entity_id) REFERENCES entities(uuid),
  FOREIGN KEY (parent_id) REFERENCES annotations(uuid)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
``` 
A migration script (stored procedure), would iterate each Entity row containing a (JSON) `annotations` object, extracting 
annotations and storing their types [`reply`, `like`, `share`, ...] into  `entity_subtype`. So for example a reply could be stored as `Idno\\Annotations\\Reply`.

I'm proposing 2 new columns here, `entity_id` as a foreign key linking to the Entity object it relates to, and `parent_id` for linking to another annotation, allowing for threads with depth.

Where an Annotation originates from a remote site (webmentions, activitypub), I would create a new `Idno\\Entity\\User` storing the minimum needed/available info [`owner_name`, `owner_url`, `owner_icon`].

Any thoughts, concerns, suggestions?

benwerd:
@mediaformat I think this looks great, and I very much see the need to make this migration. But I do have a big caveat:

Because not everyone is using the same database engine, I don't think migration should be a stored procedure. Or we should test the stored procedure for each of MySQL and Postgres, and create a command line tool for SQLite and MongoDB.

I think, actually, the latter is preferable.

It's imperative that the revised annotations methods use the database abstraction layer (presumably adding inherited methods for each db engine) rather than going straight to the SQL.

So here's what I'd suggest:

- You continue down the stored procedure road for MySQL and Postgres.
- I take the schema you've developed and adapt it for MongoDB and SQLite, with a command line update script.
- We collaborate on the annotation methods.

Does that seem reasonable?

mediaformat:
@benwerd Totally reasonable!

Idno/Common/Entity.php
```diff
@@ -2670,6 +2671,12 @@ function getAnnotationSubtype($uuid)
          */
         function getAnnotation($uuid)
         {
+            // Prioritize Annotation methods
+            $annotation = Annotation::getByUUID($uuid);
```
mediaformat, line 2675: Overriding the internals here allow us to keep using the same `getAnnotation()` calls used in templates.

Idno/Entities/Annotation.php
```diff
@@ -0,0 +1,98 @@
...
+         * @param array $search List of filter terms (default: none)
+         * @param array $fields List of fields to return (default: all)
+         * @param int $limit Number of items to return (default: 10)
+         * @param int $offset Number of items to skip (default: 0)
+         * @param array $readGroups Which ACL groups should we check? (default: everything the user can see)
+         * @return array
+         */
+        static function getFromX($class, $search = array(), $fields = array(), $limit = 10, $offset = 0, $readGroups = [])
+        {
+            $result = \Idno\Core\Idno::site()->db()->getObject($search['uuid'], static::$retrieve_collection);
```
mediaformat, line 91: The result here is a different object than was previously received, notably missing are the `owner_name`, `owner_url`, `owner_image`.

I think it would make sense to use `JOIN`s here to recreate the expected structure.

mediaformat (commented):
@benwerd Started on the Annotation methods, appreciate any early feedback.

